PR #2, opened 2020-06-22
More boilerplate

Awjin:


tool/embedded-protocol.js
```diff
@@ -0,0 +1,70 @@
...
+const execSync = require('child_process').execSync;
+const https = require('https');
+
+const PROTO_NAME = 'embedded_sass.proto';
+
+/**
+ * Downloads the latest version of the Embedded Protocol into the given
+ * directory.
+ */
+function downloadProto(dir, callback) {
```
nex3, line 15: You could probably clean this up by using https://www.npmjs.com/package/node-fetch with `async`/`await` (at the cost of adding another dev dependency).

tool/post-install.js
```diff
@@ -0,0 +1,17 @@
...
+// Create the vendor directory.
+const dir = 'dist/lib/src/vendor';
+mkdir(dir, {recursive: true}, error => {
+  if (error) {
+    throw Error(`Could not create vendor directory.\n${error.message}`);
+  }
+});
+
+// Download and compile the Embedded Sass Protocol.
+protocol.downloadProto(dir, () => protocol.writePbjs(dir));
```
nex3, line 17: Why not write the declaration file as well?

tool/post-install.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+const mkdir = require('fs').mkdir;
+const protocol = require('./embedded-protocol');
+
+// Create the vendor directory.
+const dir = 'dist/lib/src/vendor';
+mkdir(dir, {recursive: true}, error => {
```
nex3, line 10: Consider using the `fs.promises` API along with top-level `await`.

tool/embedded-protocol.ts
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {execSync} from 'child_process';
+import {promises} from 'fs';
```
nex3, line 6:
```suggestion
import {promises as fs} from 'fs';
```

This'll get you cleaner-looking invocation sites :).

Awjin: Thanks!

PR #3, opened 2020-07-13
Wrapper for dart-sass-embedded executable.

Awjin:
Abstracts away invoking the executable.

The tests here are just smoke tests (mostly to ensure this works across Node environments). Compiler behavior will be covered later on by testing the units that send data to the executable, and also through integration tests.

### DIFF
--- a/lib/src/embedded/compiler/process.test.ts
+++ b/lib/src/embedded/compiler/process.test.ts
@@ -1,0 +1,73 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {EmbeddedProcess} from './process';
+import {take} from 'rxjs/operators';
+
+describe('embedded process invocation smoke test', () => {
+  it('spins up child process', () => {
+    const process = new EmbeddedProcess();
+    process.close();
+  });
+
+  it('writes to stdin', () => {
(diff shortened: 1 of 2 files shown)


lib/src/embedded/compiler/process.test.ts
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
```
nex3, line 1: Up to you, but I generally find it clearer if tests live in a separate top-level directory.

lib/src/embedded/compiler/process.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export class EmbeddedProcess {
+  private readonly process = spawn(
+    resolve(__dirname, '../../vendor/dart-sass-embedded'),
+    {
+      windowsHide: true,
+    }
+  );
+
+  private readonly _stdout$ = new Subject<Buffer>();
```
nex3, line 22: I believe using `_` as a prefix violates the Google TS style guide, which we should try to follow here. Maybe `rawStdout$` instead?

Also, it's a good idea to document why this is different than the public property.

lib/src/embedded/compiler/process.test.ts
Awjin: I was unsure of this too. The Google typescript style guide recommends putting tests next to the source. If it starts getting out of hand I'll move them into separate directories.

Issue #4, opened 2020-07-15
Invoke and handle the Embedded Sass Compiler


Awjin:
The host is responsible for running the embedded compiler and communicating with it via the Embedded Sass Protocol. The host should invoke the compiler process and encode/decode messages to it appropriately.

PR #5, opened 2020-07-17
Packet transformer.

Awjin:
Transforms raw binary data (e.g. the buffers from the EmbeddedProcess's
stdio) into packets of delimited length.

See https://github.com/sass/embedded-host-node/issues/4

### DIFF
--- a/lib/src/embedded/compiler/packet-transformer.test.ts
+++ b/lib/src/embedded/compiler/packet-transformer.test.ts
@@ -69,14 +69,14 @@ describe('decode', () => {
     it('decodes a single chunk', async () => {
       toDecode$.next(Buffer.from([0, 0, 0, 0]));
       transformer.close();
-      expect(await decoded).toEqual([]);
+      expect(await decoded).toEqual([Buffer.from([])]);
     });
 
     it('decodes multiple chunks', async () => {
       toDecode$.next(Buffer.from([0, 0]));
       toDecode$.next(Buffer.from([0, 0]));
       transformer.close();
-      expect(await decoded).toEqual([]);
+      expect(await decoded).toEqual([Buffer.from([])]);
     });
 
     it('decodes one chunk per byte', async () => {
@@ -85,13 +85,13 @@ describe('decode', () => {
       toDecode$.next(Buffer.from([0]));
       toDecode$.next(Buffer.from([0]));
       transformer.close();
-      expect(await decoded).toEqual([]);
+      expect(await decoded).toEqual([Buffer.from([])]);
     });
 
     it('decodes a chunk that contains more data', async () => {
       toDecode$.next(Buffer.from([0, 0, 0, 0, 1, 0, 0, 0, 100]));
       transformer.close();
-      expect(await decoded).toEqual([Buffer.from([100])]);
+      expect(await decoded).toEqual([Buffer.from([]), Buffer.from([100])]);
     });
   });
 
--- a/lib/src/embedded/compiler/packet-transformer.test.ts
+++ b/lib/src/embedded/compiler/packet-transformer.test.ts
@@ -1,0 +1,168 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {PacketTransformer} from './packet-transformer';
+import {Subject, Observable} from 'rxjs';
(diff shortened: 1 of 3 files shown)


lib/src/embedded/compiler/packet-transformer.ts
```diff
@@ -0,0 +1,143 @@
...
+ *   | 0 1 2 3 | 4 5 6 7 ...
+ *   +---------+------------- ...
+ *   | HEADER  | PAYLOAD
+ *   +---------+------------- ...
+ * and emits the payload of each packet.
+ *
+ * Encodes payloads by attaching a header that describes the payload's length.
+ */
+export class PacketTransformer {
+  private readonly _read$ = new Subject<Buffer>();
```
Awjin, line 22: TODO: The underscore violates the style guide, but TBH this variable only exists to hide the Subject from public consumption. Need to find a name that captures this intent

Awjin: TO DONE

lib/src/embedded/compiler/packet-transformer.ts
```diff
@@ -0,0 +1,146 @@
...
+      bytesWritten = writeBuffer(source, this.header, this.headerOffset);
+      this.headerOffset += bytesWritten;
+      if (this.headerOffset < this.header.length) return bytesWritten;
+    }
+
+    if (!this.payload) {
+      let payloadLength = 0;
+      for (let i = 0; i < Packet.headerByteSize; i++) {
+        payloadLength += this.header[i] * Math.pow(256, i);
+      }
```
nex3, line 120: You can do this automatically using [`Buffer.readUInt32LE()`](https://nodejs.org/api/buffer.html#buffer_buf_readuint32le_offset).

lib/src/embedded/compiler/packet-transformer.ts
```diff
@@ -0,0 +1,146 @@
...
+    );
+  }
+
+  /**
+   * Takes arbitrary binary input and slots it into the header and payload
+   * appropriately. Returns the number of bytes that were written into the
+   * packet. This method can be called repeatedly, incrementally building
+   * up the packet until it is complete.
+   */
+  write(source: Buffer): number {
```
nex3, line 107: Consider adding an assertion here that no more data is written once the payload had been fully read, just to be sure future code changes don't accidentally start re-using `Packet`s in ways that break them.

lib/src/embedded/compiler/packet-transformer.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      toDecode$.next(Buffer.from([0]));
+      toDecode$.next(Buffer.from([0]));
+      transformer.close();
+      expect(await decoded).toEqual([]);
+    });
+
+    it('decodes a chunk that contains more data', async () => {
+      toDecode$.next(Buffer.from([0, 0, 0, 0, 1, 0, 0, 0, 100]));
+      transformer.close();
+      expect(await decoded).toEqual([Buffer.from([100])]);
```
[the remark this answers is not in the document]
Awjin: The transformer filters out empty packets so that consumers don't have to deal with handling them (currently undocumented behavior). Does that approach make sense?

lib/src/embedded/compiler/packet-transformer.ts
Awjin: Ah, good point!

lib/src/embedded/compiler/packet-transformer.test.ts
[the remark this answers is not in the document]
nex3: I would expect it to pass on empty packets to the consumer. They're definitely not valid protocol buffers, but they should be covered by the [error handling](https://github.com/sass/embedded-protocol#error-handling) part of the spec, as "data that can't be parsed".

lib/src/embedded/compiler/packet-transformer.ts
```diff
@@ -0,0 +1,159 @@
...
+  private encode(payload: Buffer) {
+    const packet = Buffer.alloc(Packet.headerByteSize + payload.length);
+    packet.writeInt32LE(payload.length, 0);
+    packet.set(payload, Packet.headerByteSize);
+    this.rawWrite$.next(packet);
+  }
+
+  /**
+   * Cleans up all Observables.
+   */
```
nex3, line 68:
```suggestion
  /** Cleans up all Observables. */
```

Style nit

PR #6, opened 2020-07-27
Message transformer.

Awjin:
Transforms buffers into Inbound/OutboundMessages.

See https://github.com/sass/embedded-host-node/issues/4.


lib/src/embedded/compiler/message-transformer.ts
```diff
@@ -0,0 +1,222 @@
...
+  } catch (error) {
+    throw protocolError(ProtocolError.ErrorType.PARSE, 'Invalid buffer');
+  }
+
+  // Ensure that the InboundMessage contains a valid message, and that all
+  // mandatory fields are populated.
+  switch (message.getMessageCase()) {
+    case OutboundMessage.MessageCase.COMPILERESPONSE: {
+      const response = message.getCompileresponse()!;
+      if (!response.getId()) {
```
nex3, line 106: There's some tricky stuff going on here. Scalar values in protocol buffers (including ints and even strings) are never null. If the message doesn't include an explicit value, the message will use a [default value](https://developers.google.com/protocol-buffers/docs/proto3#default). This means there's no way to distinguish between a message passing an explicit `0` value for an ID and leaving the ID out entirely. Since the embedded protocol allows 0 as an ID, this effectively means that no message is inherently invalid because it doesn't include an ID. The good news is that this means you have to do a lot less manual checking in this method!

Also, be careful of relying on truthiness in JavaScript—it can behave in unintuitive ways. For example, `0` is falsey, so this will currently throw an error if the ID is set to `0` (explicitly or implicitly). One of those annoying corner cases!

lib/src/embedded/compiler/message-transformer.ts
```diff
@@ -0,0 +1,222 @@
...
+        throw protocolError(
+          ProtocolError.ErrorType.PARAMS,
+          'OutboundMessage.CompileResponse.id is not set'
+        );
+      }
+      if (
+        response.getResultCase() ===
+        OutboundMessage.CompileResponse.ResultCase.RESULT_NOT_SET
+      ) {
+        throw protocolError(
```
nex3, line 116: Shouldn't this have the message ID associated with it?

lib/src/embedded/compiler/message-transformer.test.ts
```diff
@@ -0,0 +1,275 @@
...
+        8,
+        97,
+        32,
+        123,
+        98,
+        58,
+        32,
+        99,
+        125,
+      ]),
```
nex3, line 52: Rather than manually writing out the encoded message (which could potentially change with protobuf library changes), maybe just assert that it's equal to `message.serializeBinary()`. It's safe to assume that the protobuf library's serialization is correct and doesn't need to be re-tested here.

lib/src/embedded/compiler/message-transformer.test.ts
```diff
@@ -0,0 +1,275 @@
...
+        8,
+        97,
+        32,
+        123,
+        98,
+        58,
+        32,
+        99,
+        125,
+      ])
```
nex3, line 93: Similarly to above, it's probably simpler to just construct a protobuf object and call `serializeBinary()` on it rather than manually encoding the bytes.

lib/src/embedded/compiler/message-transformer.ts
```diff
@@ -46,177 +35,87 @@ export class MessageTransformer {
...
   // Decodes a buffer into an OutboundMessage.
   private decode(buffer: Buffer) {
-    let message;
     try {
-      message = toMessage(buffer);
+      const message = toMessage(buffer);
+      this.readInternal$.next(message);
     } catch (error) {
-      this.protocolErrorInternal$.next(error);
+      this.errorInternal$.next(error);
```
nex3, line 42: Should this close upon encountering an error? Given that, should the error stream just be a promise instead?

nex3 (approved):

lib/src/embedded/compiler/message-transformer.ts
Awjin: Yes, eventually—I'm planning to have the handler who instantiates the message transformer also handle closing it.

Awjin: I do think it's advantageous to keep it as an Observable for as long as possible, until we're ready to resolve and return its value to the API call

PR #7, opened 2020-09-30
Adds a compiler facade.

Awjin:
This integrates the process, packet transformer, and message
transformer.

### DIFF
--- a/lib/src/embedded/compiler.test.ts
+++ b/lib/src/embedded/compiler.test.ts
@@ -2,12 +2,12 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-import {Compiler} from './compiler';
+import {EmbeddedCompiler} from './compiler';
 import {InboundMessage} from '../vendor/embedded_sass_pb';
 
-describe('compiler', () => {
-  let compiler: Compiler;
-  beforeEach(() => (compiler = new Compiler()));
+describe('embedded compiler', () => {
+  let compiler: EmbeddedCompiler;
+  beforeEach(() => (compiler = new EmbeddedCompiler()));
   afterEach(() => compiler.close());
 
   it('compiles valid css input', async done => {
@@ -41,9 +41,9 @@ describe('compiler', () => {
     compiler.write$.next(new InboundMessage());
     compiler.read$.subscribe(
       // Response callback
-      () => {},
+      () => fail(),
       // Error callback
-      () => {},
+      () => fail(),
       // Completion callback
       () => done()
     );
--- a/lib/src/embedded/compiler.test.ts
+++ b/lib/src/embedded/compiler.test.ts
@@ -1,0 +1,51 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {Compiler} from './compiler';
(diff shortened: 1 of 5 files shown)


Awjin:
The diffs for `message-transformer.test.ts` and `packet-transformer.test.ts` are a bit silly. The only changes I made were wrapping the whole spec in a describe block to make the terminal output more readable when running tests.

lib/src/embedded/compiler.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Takes the compiler's raw stdin and stdout and transforms them into public
+ * Observables of InboundMessages (write$) and OutboundMessages (read$). Also
+ * exposes an Observable (error$) that emits if a ProtocolError occurs while
+ * communicating with the compiler.
+ *
+ * If there is a ProtocolError, this shuts down the compiler process and cleans
+ * up all Observables.
+ */
+export class Compiler {
```
nex3, line 23: I suspect at some point we're going to want to expose a public class called `Compiler` to address https://github.com/sass/sass/issues/2745, so it might be worth finding a different name for this now so we don't have to rename it later. Maybe `Host`?

lib/src/embedded/compiler.ts
```diff
@@ -0,0 +1,79 @@
...
+  // The Node child process that invokes the Embedded Sass Compiler.
+  private readonly embeddedProcess = new EmbeddedProcess();
+
+  // Transforms the embedded process's into delineated buffers.
```
nex3, line 27: Nit

```suggestion
  // Transforms the embedded process's IO streams into delineated buffers.
```

lib/src/embedded/compiler.ts
```diff
@@ -0,0 +1,79 @@
...
+    )
+  );
+
+  constructor() {
+    // Pass the embedded process's stderr messages to the main process's stderr.
+    this.embeddedProcess.stderr$.subscribe(buffer => {
+      process.stderr.write(buffer);
+    });
+
+    // Shutdown the compiler if we detect a ProtocolError or the embedded
```
nex3, line 66: Nit

```suggestion
    // Shut down the compiler if we detect a ProtocolError or the embedded
```

"Shutdown" is a noun, "shut down" is a verb :smiley:.

lib/src/embedded/compiler.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      );
+      done();
+    });
+  });
+
+  it('shuts itself down when a ProtocolError occurs', async done => {
+    compiler.write$.next(new InboundMessage());
+    compiler.read$.subscribe(
+      // Response callback
+      () => {},
```
nex3, line 44: Is there an easy way to verify that the response and error callbacks are never called?

nex3 (approved):

lib/src/embedded/compiler.test.ts
Awjin: I looked pretty seriously into marble testing for rxjs, but it doesn't work since the compiler's `read$` observable completes in a `process.nextTick()`, and there's no way for the marble test to execute after the tick. It's a shame, bc marble testing would be good for this kind of test.

Instead, I made the response and error branches manually fail the jasmine test.

lib/src/embedded/compiler.ts
Awjin: Good point, thanks for calling this out

PR #8, opened 2020-10-14
Adds a dispatcher.

Awjin:


Awjin:
@nex3 I added the changes you suggested, wdyt?

lib/src/embedded/compiler.ts
```diff
@@ -118,10 +84,132 @@ export class EmbeddedCompiler {
       .subscribe(() => process.nextTick(() => this.close()));
   }
 
+  /**
+   * Dispatches a compile request. Returns a promise that resolves when a
+   * compile response with a matching ID is received from the embedded compiler.
+   */
+  compile(request: InboundMessage.CompileRequest) {
```
nex3: Please add return type annotations for all methods.

lib/src/embedded/compiler.ts
```diff
@@ -118,10 +84,132 @@ export class EmbeddedCompiler {
...
+    const message = new InboundMessage();
+    message.setCompilerequest(request);
+
+    this.recordRequest(id);
+    this.messageTransformer.write$.next(message);
+
+    return this.messageTransformer.read$
+      .pipe(
+        filter(message => {
+          return (
```
nex3: Nit: unnecessary parentheses

lib/src/embedded/compiler.ts
```diff
@@ -118,10 +84,132 @@ export class EmbeddedCompiler {
...
+        take(1),
+        takeUntil(this.error$)
+      )
+      .toPromise();
+  }
+
+  listen(
+    messageType: OutboundMessage.MessageCase,
+    processRequest: (request: Request) => Response
+  ) {
```
nex3: Rather than having a generic public `listen()` method, I think it would be better to expose separate methods for the various types of requests that can be listened for, such as:

```ts
onImportRequest(
  (request: OutboundMessage.ImportRequest) => Promise<InboundMessage.ImportResponse>
): Subscription
```

That way, you have a type-system-level guarantee that the caller is not only returning a response, but returning the appropriate response type.

(I also think you'll want to make the handler asynchronous since you'll potentially be calling out to async importer and function implementations.)

lib/src/embedded/compiler.ts
```diff
@@ -7,8 +7,21 @@ import {EmbeddedProcess} from './compiler/process';
...
-import {filter, first, map} from 'rxjs/operators';
-import {Dispatcher} from './dispatcher';
+import {filter, first, map, take, takeUntil, tap} from 'rxjs/operators';
+
+type Request =
+  | InboundMessage.CompileRequest
+  | OutboundMessage.ImportRequest
+  | OutboundMessage.FileImportRequest
+  | OutboundMessage.CanonicalizeRequest
+  | OutboundMessage.FunctionCallRequest;
```
nex3: Why is it useful to group inbound and outbound requests under the same type?

lib/src/embedded/compiler.ts
```diff
@@ -118,10 +84,132 @@ export class EmbeddedCompiler {
...
+
+  listen(
+    messageType: OutboundMessage.MessageCase,
+    processRequest: (request: Request) => Response
+  ) {
+    this.messageTransformer.read$
+      .pipe(
+        filter(message => message.getMessageCase() === messageType),
+        map(message => getRequest(message)),
+        tap(request => this.recordRequest(request.getId())),
```
nex3: Do you need to record the ID here? You're allowed to have your `InboundRequest` IDs overlap with `OutboundRequest` IDs.

lib/src/embedded/compiler.ts
```diff
@@ -118,10 +84,132 @@ export class EmbeddedCompiler {
...
+        map(request => processRequest(request)),
```
nex3: Consider automatically adding the appropriate ID to the response object, so you don't need to duplicate that logic across each caller.

nex3 (changes requested):

lib/src/embedded/compiler.ts
Awjin: Thanks, that's a better solution.

lib/src/embedded/compiler.ts
Awjin: It's not :) This was a mistake corrected in a newer commit.

lib/src/embedded/compiler.ts
Awjin: I'm recording it here in case the compiler sends more than one request with with the same request ID. Is that something we want to protect against?

nex3: I think that's reasonable to check, but I'd probably use a different set for inbound IDs and outbound IDs—otherwise, a reader (like me!) is liable to get confused about why you're tracking outbound IDs at all.

Awjin:
@nex3 I'm closing this in favor of https://github.com/sass/embedded-host-node/pull/9. That one incorporates all the feedback here, but is cleaned up and can be reviewed commit-by-commit.

PR #16, opened 2020-11-11
Travis integration

Awjin:
Closes https://github.com/sass/embedded-host-node/issues/11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,20 +15,9 @@ os:
 - osx
 - windows
 
-env:
-  global:
-  - PROTOBUF_VERSION=3.13.0
-  - PATH="$HOME/protoc/bin:$PATH"
-
 cache: npm
 
 install:
-- if [[ "$TRAVIS_OS_NAME" = windows ]]; then
-    curl -Lo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOBUF_VERSION}/protoc-${PROTOBUF_VERSION}-win64.zip;
-  else
-    curl -Lo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOBUF_VERSION}/protoc-${PROTOBUF_VERSION}-${TRAVIS_OS_NAME}-x86_64.zip;
-  fi
-- unzip -d "$HOME/protoc" protoc.zip
 - npm install
 - npm run init
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,42 @@
+# Only building master means that we don't run two builds for each pull request.
+branches:
+  only:
+  - master
+
+language: node_js
+
+env:
+  global:
+  - PROTOBUF_VERSION=3.13.0
+  - PATH="$HOME/protoc/bin:$PATH"
+
+cache: npm
+
+install:
+- if [[ "$TRAVIS_OS_NAME" = windows ]]; then
+    curl -Lo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOBUF_VERSION}/protoc-${PROTOBUF_VERSION}-win64.zip;
+  else
+    curl -Lo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOBUF_VERSION}/protoc-${PROTOBUF_VERSION}-${TRAVIS_OS_NAME}-x86_64.zip;
+  fi
+- unzip -d "$HOME/protoc" protoc.zip
+- npm install
+- npm run init
+
+script:
+- npm test
+
+jobs:
+  include:
+  ## Testing
+  - node_js: "14"
+    os: linux
+  - node_js: "14"
+    os: osx
+  - node_js: "13"
+    os: linux
+  - node_js: "13"
+    os: osx
+  - node_js: "12"
+    os: linux
+  - node_js: "12"
+    os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,16 @@ branches:
 
 language: node_js
 
+node_js:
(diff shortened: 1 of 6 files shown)


tool/embedded-binaries.ts
```diff
@@ -61,20 +61,26 @@ export async function getDartSassEmbedded(outPath: string): Promise<void> {
...
     // API exposes only the latest *non*-pre-release versions.
     // https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-the-latest-release
-    let response = await fetch(
-      'https://api.github.com/repos/sass/dart-sass-embedded/releases'
-    );
-    if (!response.ok) throw Error(response.statusText);
-    const latestRelease = JSON.parse(await response.text())[0];
+    //
+    // Mock the latest release version for now. Travis can't call the Github API
+    // because Github seems to be throttling all calls from the osx instance.
```
Awjin, line 66: Is there a way to set up a repo token to get around this? It seems like all anonymous requests by Travis's osx instances to the Github API are using the same IP address, or otherwise saturating the API.

(Keep in mind that this script needs to be run by both Travis and any contributors developing locally.)

nex3: I've added a `GITHUB_AUTH` environment variable that's visible on Travis and has SassBot's authorization token. If you add that to Basic Auth with the username "sassbot" that should work. Locally, users will be using different IPs so they're unlikely to run up against rate-limiting.

[This is the logic we use for GitHub requests in Dart Sass](https://github.com/sass/dart-sass/blob/4f1f5c9de80499ab2f4bface7dbda2d3ef50a41e/tool/travis/sass-spec-ref.sh#L25-L32), for reference.

nex3 (changes requested):

.travis.yml
```diff
@@ -0,0 +1,42 @@
...
+  - node_js: "14"
+    os: osx
+  - node_js: "13"
+    os: linux
+  - node_js: "13"
+    os: osx
+  - node_js: "12"
+    os: linux
+  - node_js: "12"
+    os: osx
```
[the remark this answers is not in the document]
Awjin: ah, thanks

.travis.yml
```diff
@@ -0,0 +1,42 @@
...
+- npm run init
+
+script:
+- npm test
+
+jobs:
+  include:
+  ## Testing
+  - node_js: "14"
+    os: linux
```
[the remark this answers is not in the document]
Awjin: Whoops, forgot to add this back in after trying to get unix done. Fixed

.travis.yml
```diff
@@ -0,0 +1,42 @@
...
+  - node_js: "14"
+    os: osx
+  - node_js: "13"
```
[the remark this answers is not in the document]
Awjin: Good to know, thanks

tool/embedded-binaries.ts
```diff
@@ -19,7 +20,10 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
   console.log('Downloading Embedded Sass Protocol.');
   try {
     const response = await fetch(
-      `https://raw.githubusercontent.com/sass/embedded-protocol/master/${protoName}`
+      `https://raw.githubusercontent.com/sass/embedded-protocol/master/${protoName}`,
+      {
+        redirect: 'follow' as RequestRedirect,
```
nex3, line 25: I think this `as` shouldn't be necessary... `as` asserts that an expression of a broader type is in fact a narrower type, but `'follow'` is statically known to be a member of `RequestRedirect`. Similar comment below.

tool/embedded-binaries.ts
```diff
@@ -51,46 +59,73 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
-    //
-    // TODO(awjin): Once dart-sass-embedded is no longer under pre-release, call
-    // the Github API instead of manually piecing together the asset URL. The
-    // API exposes only the latest *non*-pre-release versions.
-    // https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-the-latest-release
-    let response = await fetch(
-      'https://api.github.com/repos/sass/dart-sass-embedded/releases'
+    const fetchOptions = {
+      redirect: 'follow' as RequestRedirect,
+    } as RequestInit;
```
nex3, line 76:
```suggestion
    const fetchOptions: RequestInit = {
      redirect: 'follow' as RequestRedirect,
    };
```

I believe this is a more idomatic way to add a type restriction to a variable (although I wouldn't call myself an expert here).

tool/embedded-binaries.ts
```diff
@@ -51,46 +59,73 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
+    if (process.env.TRAVIS === 'true') {
```
nex3, line 77: I'd also verify that `process.env.GITHUB_AUTH` is set. If non-team-members submit PRs, it won't be.

.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+
+env:
+  global:
+  - PROTOBUF_VERSION=3.13.0
+  - PATH="$HOME/protoc/bin:$PATH"
+
+cache: npm
+
+install:
+- if [[ "$TRAVIS_OS_NAME" = windows ]]; then
```
nex3, line 26: Can you add a comment here explaining why we can't use the protoc npm package?

nex3 (changes requested):
Thank you for slogging through all the Travis config headaches!

.travis.yml
Awjin: No good reason, it makes sense to do this through npm.

nex3 (approved):

Issue #13, opened 2020-11-03
Implement asynchronous API `render()`


Awjin:
This should replicate the current Sass JS API call. This work can be split up by first releasing a stripped down version of `render` that supports core behavior, then gradually adding more advanced support for custom importers and functions:

- [x] basic compile functionality with sourcemaps
- [x] syntax and style control
- [x] custom functions
- [x] custom importers

dickeylth:
Are we making any progress on `custom importers` support?

Awjin:
This issue is outdated. Importers are now supported.

PR #17, opened 2020-11-17
Compile function

Awjin:
Adds a `compile` function that will back the `render` API (and eventually `renderSync` as well as the new JS API methods).

### DIFF
--- a/lib/src/compile.ts
+++ b/lib/src/compile.ts
@@ -1,0 +1,95 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {URL} from 'url';
+
+import {EmbeddedCompiler} from './embedded/compiler';
+import {Dispatcher} from './embedded/dispatcher';
+import {MessageTransformer} from './embedded/message-transformer';
+import {PacketTransformer} from './embedded/packet-transformer';
+import {InboundMessage, OutboundMessage} from './vendor/embedded_sass_pb';
+
+/**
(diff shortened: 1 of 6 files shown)


spec/helpers/utils.ts
```diff
@@ -22,3 +28,88 @@ export function expectError<T>(
     () => fail('expected error')
   );
 }
+
+/**
+ * Asserts that the `actual` path is equal to the `expected` one, accounting for
+ * OS differences.
+ */
+export function expectEqualPaths(actual: string, expected: string) {
```
Awjin, line 36: This mirrors the behavior I saw from dart-sass. Does dart-sass lowercase all filenames on Windows?

nex3: Yes, the canonicalization process works around https://github.com/dart-lang/sdk/issues/35370 by eagerly converting all path names to lowercase to avoid case-sensitivity issues.

lib/src/compile.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/**
+ * Spins up a compiler, then sends it a compile request. Returns a promise that
+ * resolves with the compilation result. After resolving, shuts down the
+ * compiler.
+ */
+export function compile(
+  request: InboundMessage.CompileRequest
+): Promise<OutboundMessage.CompileResponse> {
```
nex3, line 55: I'm surprised that this function takes and returns protobufs, rather than having a more JS-idiomatic API. Once you add support for importers, it won't be easy to decouple the protobuf creation with the event handling, because the importers are all addressed by ID.

Awjin:
Closing this in favor of https://github.com/sass/embedded-host-node/pull/19

PR #18, opened 2020-11-19
Add exception class and helpers.

Awjin:


lib/src/exception/exception.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    this.span = options.span;
+
+    if (options.stack) {
+      this.stack = options.stack;
+    } else {
+      Error.captureStackTrace(this, this.constructor);
+    }
+  }
+}
```
nex3, line 23: This should have a `toString()` that produces a human-readable error with the source span highlighted appropriately. You can add a TODO and file an issue for this for now, but it should be in place before we launch the beta.

lib/src/utils.ts
```diff
@@ -2,4 +2,30 @@
...
+import * as proto from './vendor/embedded_sass_pb';
+import {SourceLocation} from './exception/location';
+import {SourceSpan} from './exception/span';
+
 export type PromiseOr<T> = T | Promise<T>;
+
+export function deprotifySourceSpan(span: proto.SourceSpan): SourceSpan {
+  return {
+    text: span.getText(),
+    start: deprotifySourceLocation(span.getStart()!),
```
nex3, line 16: We should explicitly verify that these fields aren't null ("the endpoint that receives [a message] must reject that message if [a mandatory field] is not set") and throw a user-facing error if they are, since it constitutes a protocol violation.

lib/src/exception/exception.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+describe('SassException', () => {
+  it('has the correct type', () => {
+    const error = new SassException({message: ''});
+
+    try {
+      throw error;
+    } catch (error) {
+      expect(error instanceof SassException).toBe(true);
+    }
```
nex3, line 18: I'm pretty sure this much is guaranteed by the JS language, so you probably don't need to test it explicitly.

lib/src/exception/exception.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {SourceSpan} from './span';
+
+export class SassException extends Error {
+  span?: SourceSpan;
```
nex3, line 8: This should be `readonly`.

lib/src/exception/exception.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  constructor(options: {message: string; span?: SourceSpan; stack?: string}) {
```
nex3, line 10: I'd make `message` a positional argument, both to match `new Error()` and because it's always required and unlikely to be ambiguous with the other arguments.

nex3 (changes requested):

Awjin:
@nex3 I still have to test the deprotification class methods, but does this look better?

nex3 (approved):

Awjin:
Changed my mind here. I think it's better to keep the deprotification methods out of the classes, since we'll need protification methods that act on Sass Values that shouldn't be exposed as part of those value classes.

nex3 (approved):

PR #19, opened 2020-11-24
Compile

Awjin:
Adds a `compile` function that will back the `render` API (and eventually `renderSync` as well as the new JS API methods).

### DIFF
--- a/lib/src/compile.test.ts
+++ b/lib/src/compile.test.ts
@@ -1,0 +1,213 @@
+// Copyright 2020 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {promises as fs} from 'fs';
+import {resolve} from 'path';
+import {fileURLToPath, URL} from 'url';
+
+import {compile, compileString} from './compile';
+import {expectEqualPaths} from '../../spec/helpers/utils';
+import {RawSourceMap, SourceMapConsumer} from 'source-map';
+
+describe('compile', () => {
+  describe('success', () => {
+    describe('input', () => {
+      it('compiles an SCSS string by default', async () => {
+        expect(await compileString({source: 'a {b: c}'})).toBe(
+          'a {\n  b: c;\n}'
+        );
+      });
+
+      // TODO(awjin): compiles an SCSS string explicitly
+      // TODO(awjin): compiles an indented syntax string
+      // TODO(awjin): compiles a plain CSS string
+
+      it('compiles an absolute path', async () => {
+        const path = `${resolve('test.scss')}`;
+        await fs.writeFile(path, 'a {b: c}');
+
+        expect(await compile({path})).toBe('a {\n  b: c;\n}');
+        await fs.unlink(path);
+      });
+
+      it('compiles a relative path', async () => {
+        const path = 'test.scss';
+        await fs.writeFile(path, 'a {b: c}');
+
+        expect(await compile({path})).toBe('a {\n  b: c;\n}');
+        await fs.unlink(path);
+      });
+    });
+
+    describe('output', () => {
+      it('outputs in expanded mode by default', async () => {
+        expect(await compileString({source: 'a {b: c}'})).toBe(
+          'a {\n  b: c;\n}'
+        );
+      });
+
+      // TODO(awjin): outputs in expanded mode explicitly
+      // TODO(awjin): outputs in compressed mode
+      // TODO(awjin): outputs in expanded mode when nested mode is passed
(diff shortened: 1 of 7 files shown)


lib/src/compile.ts
```diff
@@ -0,0 +1,163 @@
...
+  // TODO(awjin): Create logger, importer, function registries.
+
+  const request = newCompileStringRequest({
+    source: options.source,
+    sourceMap: !!options.sourceMap,
+    url: options.url,
+  });
+
+  const response = await compileRequest(request);
+  if (options.sourceMap && response.sourceMap) {
```
nex3, line 56: Shouldn't it be an error if `options.sourceMap` is set and `response.sourceMap` is empty?

lib/src/compile.test.ts
```diff
@@ -0,0 +1,213 @@
...
+          'a {\n  b: c;\n}'
+        );
+      });
+
+      // TODO(awjin): compiles an SCSS string explicitly
+      // TODO(awjin): compiles an indented syntax string
+      // TODO(awjin): compiles a plain CSS string
+
+      it('compiles an absolute path', async () => {
+        const path = `${resolve('test.scss')}`;
```
nex3, line 27: I don't think you need the interpolated string here.

```suggestion
        const path = resolve('test.scss');
```

lib/src/compile.test.ts
```diff
@@ -0,0 +1,213 @@
...
+            context: 'a {b: 1px + 1em}',
+          });
+          expect(error.stack).toBe('- 1:7  root stylesheet\n');
+        }
+      });
+
+      it('fails on missing file', async () => {
+        try {
+          await compile({path: 'test.scss'});
+        } catch (error) {
```
nex3, line 129: I'd check that the span is null here, too.

nex3 (approved):
Just a few more small changes!

lib/src/compile.ts
Awjin: Yes, good catch

lib/src/compile.ts
```diff
@@ -0,0 +1,156 @@
...
+        throw Error('Canonicalize not yet implemented.');
+      },
+      handleFunctionCallRequest: () => {
+        throw Error('Custom functions not yet implemented.');
+      },
+    }
+  );
+
+  dispatcher.error$.subscribe(error => {
+    throw error;
```
nex3, line 136: I believe this will cause the error to be passed to the top-level error handler. Rather than that, we probably want to pipe it into the returned `Promise`.

nex3 (changes requested):

lib/src/compile.ts
Awjin: Since `compileRequest` is an `async` function, it implicitly wraps everything in a promise. This `throw` then fails that promise

Awjin: Nope, I'm wrong. The subscription could fail after the promise resolves. Will fix

Awjin: Although, line 139 *should* close the subscription before this promise returns. Maybe still worth unsubscribing explicitly?

nex3: Yeah, an `async` function doesn't automatically collect all errors in its lexical scope—just those that are emitted by `await`ed Promises. You'll need to essentially "race" this error against the `sendCompileRequest()` promise, so that if `dispatcher.error$` emits an error before the promise completes `compileRequest()` forwards that error but if it doesn't the stream gets unsubscribed.

Awjin: Sigh I was massively overthinking this. The dispatcher already wraps up the error into the promise that it returns. I just have to handle it properly with a `finally` clause.

lib/src/compile.ts
```diff
@@ -132,25 +132,24 @@ async function compileRequest(
...
-  const response = await dispatcher.sendCompileRequest(request);
-  embeddedCompiler.close();
-
-  if (response.getSuccess()) {
-    const success = response.getSuccess()!;
-    const sourceMap = success.getSourceMap();
-    if (request.getSourceMap() && sourceMap === undefined) {
-      throw Error('Compiler did not provide sourceMap.');
+  try {
+    const response = await dispatcher.sendCompileRequest(request);
```
Awjin, line 136: @nex3 There is one more edge case here. If the dispatcher's underlying steam closes due to an error *before* sending this request, this `await` will hang. In this case, would it be better for the dispatcher to just resolve the `sendCompileRequest` promise, or reject it with an error?

nex3: It should reject it with an error—a promise should only resolve successfully if its designated task has completed and it has a useful value to return.

Awjin: Right, done.

nex3 (approved):

PR #21, opened 2020-12-22
Delimit messages using varints, as per sass/embedded-protocol#38

nex3:
See sass/embedded-protocol#37


Awjin (approved):
LGTM. Looks like the broken integration tests should resolve once we depend on the latest version of dart-sass-embedded

tool/embedded-binaries.ts
```diff
@@ -135,6 +211,16 @@ export async function getDartSassEmbedded(outPath: string): Promise<void> {
   }
 }
 
+// Ensures that `outPath` exists and its `sass_embedded` subdirectory does not.
+async function cleanEmbeddedDir(outPath: string): Promise<void> {
+  await fs.mkdir(outPath, {recursive: true});
+  try {
+    await fs.rmdir(p.join(outPath, 'sass_embedded'), {recursive: true});
+  } catch (_) {
+    // If the path doesn't exist yet, that's fine.kk
```
Awjin, line 220: nit: delete 'kk'

tool/embedded-binaries.ts
```diff
@@ -60,9 +61,74 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
+): Promise<void> {
+  if (release) {
+    await downloadDartSassEmbedded(outPath, version);
+  } else {
+    await buildDartSassEmbedded(outPath, version);
+  }
+}
+
+// Clones the Embedded Dart Sass repo and build its executable from source.
+async function buildDartSassEmbedded(
```
Awjin, line 86: Will this ever be called when a user downloads the npm package? My only concern is requiring Node users to depend on a Dart build step instead of just downloading the release binary.

nex3: I was only thinking about this from a development perspective. Are you expecting that this will also be run when a user installs the package? Looking at the npm docs, it looks like that's typically done in a separate postinstall script.

Awjin: I was thinking the postinstall script would also call this helper function. But, it also could make sense to just write separate logic for the postinstall

nex3: If it calls it as a helper function, it can just pass in `release: true` to ensure that the release versions are downloaded, right?

Awjin: ah, that's right

jathak (approved):

PR #23, opened 2021-01-16
Replace instances of URL with string.

Awjin:


lib/src/compile.ts
```diff
@@ -41,7 +40,7 @@ export async function compile(options: {
 export async function compileString(options: {
   source: string;
   sourceMap?: (sourceMap: RawSourceMap) => void;
-  url?: URL;
+  url?: string;
```
nex3, line 43: It might be a good idea to allow this to take a URL *or* a string.

nex3 (changes requested):

nex3 (approved):

Issue #20, opened 2020-12-02
Sass Exception needs a toString() class method


Awjin:
This should provide properly formatting, as well as color encodings for spans. See https://babeljs.io/docs/en/babel-code-frame

PR #26, opened 2021-01-21
Update SassException to properly handle span/trace.

Awjin:
Closes https://github.com/sass/embedded-host-node/issues/20

### DIFF
--- a/lib/src/compile.test.ts
+++ b/lib/src/compile.test.ts
@@ -95,7 +95,7 @@ describe('compile', () => {
             url: undefined,
             context: 'a {',
           });
-          expect(error.stack).toBe('- 1:4  root stylesheet\n');
+          expect(error.trace).toBe('- 1:4  root stylesheet\n');
         }
       });
 
@@ -119,7 +119,7 @@ describe('compile', () => {
             url: undefined,
             context: 'a {b: 1px + 1em}',
           });
-          expect(error.stack).toBe('- 1:7  root stylesheet\n');
+          expect(error.trace).toBe('- 1:7  root stylesheet\n');
         }
       });
 
@@ -131,6 +131,7 @@ describe('compile', () => {
           expect(message[0]).toBe('Cannot open file');
           expectEqualPaths(message[1], resolve('test.scss'));
           expect(error.span).toBe(undefined);
+          expect(error.trace).toBe(undefined);
         }
       });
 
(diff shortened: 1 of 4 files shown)


lib/src/exception/exception.test.ts
```diff
@@ -52,26 +52,73 @@ describe('SassException', () => {
     }
   });
 
-  it('has the default stack trace', () => {
-    const message = 'default';
-    const error = new SassException(message);
+  describe('stack trace', () => {
+    it('correctly formats the message, span, and trace', () => {
+      const message = 'sad';
+      const sass = 'wrong';
```
nex3, line 58: It's nice to try to re-use these values between the constructor and the expectation, but I think the indirection adds more confusion than it solves (especially since the messages are short and the formatting of the error depends heavily on the structure of the Sass code).

lib/src/exception/exception.ts
```diff
@@ -2,28 +2,67 @@
...
 export class SassException extends Error {
   /**
-   * @param message - The error message.
+   * @param [message] - The error message.
    * @param [span] - The source span associated with the error.
-   * @param [stack] - The stack trace associated with the error.
+   * @param [trace] - The trace associated with the error.
    */
-  constructor(message?: string, readonly span?: SourceSpan, stack?: string) {
+  constructor(message?: string, readonly span?: SourceSpan, trace?: string) {
```
nex3, line 18: One alternative to building the highlight yourself would be having the compiler provide a formatted message along with a `CompileFailure`/`LogEvent`. What do you think?

lib/src/compile.test.ts
```diff
@@ -95,7 +95,12 @@ describe('compile', () => {
             url: undefined,
             context: 'a {',
           });
-          expect(error.stack).toBe('- 1:4  root stylesheet\n');
+          expect(error.stack).toBe(`Error: expected "}".
+    ╷
+> 1 | a {
+    |    ^
+    ╵
+  - 1:4  root stylesheet`);
```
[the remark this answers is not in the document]
Awjin: This is for parity with JS errors, which contain the entire error message within `Error.stack`.

lib/src/exception/exception.ts
Awjin: I like the idea of having a formatted message provided by the compiler.

lib/src/exception/exception.test.ts
Awjin: Makes sense.

lib/src/exception/exception.ts
Awjin: If that's the case, we'd also need to tell the compiler whether to colorize the formatted message

Awjin: I updated the protocol here: https://github.com/sass/embedded-protocol/pull/49. If that looks good I can also update dart-sass-embedded.

lib/src/compile.test.ts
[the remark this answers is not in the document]
Awjin: The `Error.stack` now contains the Sass error without clobbering the JS trace

Awjin:
@nex3 PTAL; exception now use the formatted errors from dart-sass-embedded to create a useful `toString()` method.

nex3 (approved):

PR #28, opened 2021-02-05
Path helpers

Awjin:
Adds helper methods to smooth over Node's URI/path dichotomy.

### DIFF
--- a/lib/src/helpers/path.ts
+++ b/lib/src/helpers/path.ts
@@ -1,0 +1,73 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+// For the most part, Sass treats URIs and paths interchangeably, but Node does
+// not. These helper methods smooth over Node's URI/path dichotomy.
+
+import * as p from 'path';
+import {fileURLToPath, pathToFileURL, URL} from 'url';
+
+/**
+ * Returns the URL that represents `pathOrUri`.
+ *
+ * If `pathOrUri` is an absolute file path, returns a `file:` UTL.
+ * If `pathOrUri` is a URL, returns the URL.
+ * If `pathOrUri` is relative, returns a relative URI.
+ */
+export function toUrl(pathOrUri: string): string {
+  try {
+    return new URL(pathOrUri).toString();
+  } catch (error) {
+    if (p.isAbsolute(pathOrUri)) {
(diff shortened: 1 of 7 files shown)


lib/src/helpers/path.ts
```diff
@@ -0,0 +1,73 @@
...
+ * If `pathOrUri` is relative, returns a relative URI.
+ */
+export function toUrl(pathOrUri: string): string {
+  try {
+    return new URL(pathOrUri).toString();
+  } catch (error) {
+    if (p.isAbsolute(pathOrUri)) {
+      return pathToFileURL(pathOrUri).toString();
+    } else {
+      return pathOrUri;
```
nex3, line 25: Even relative paths aren't interchangeable with URLs, especially on Windows (but in edge cases also on POSIX systems). Unfortunately it looks like there's not an idiomatic way to do this transformation in Node, but you get 80% of the way there by converting backslashes to forward slashes on Windows.

nex3 (changes requested):
It's usually pretty error-prone to allow strings to represent paths *or* URLs, because there's considerable overlap and ambiguity between the two. For example, `foo%20bar` is both a valid path *and* a valid URL, but what it refers to differs based on the interpretation. Rather than trying to make both work in the same place, I recommend trying to be very clear about which one is allowed in which place and ensuring that you convert between them at the boundaries.

Awjin:
Could you explain what it means to convert at the boundaries?

nex3:
I mean the boundaries between code that deals in filesystem paths (like the `render()` API and the legacy importer API) and code that deals in URLs (like the embedded Sass protocol). When you pass a value from one to the other, you convert it between being a path and being a URL.

If it helps, you can think of them as two different types (`PathString` versus `UrlString`)—you should never pass a `UrlString` where a `PathString` is expected and vice versa, so you need functions that take one and return the other.

lib/src/helpers/path.ts
Awjin: Ah, got it. I understand the Windows case, but shouldn't a relative POSIX path be interchangeable with a relative URL?

nex3: No, for two reasons:

1. POSIX paths allow effectively any character (technically, byte sequence) to exist unescaped, while URLs only allow a subset of ASCII.
2. The `%` character in particular is a literal character in POSIX paths but an escape character in URLs, so for URLs `f%6Fo` is equivalent to `foo` but for POSIX paths they're totally different.

PR #29, opened 2021-02-09
Render fn

Awjin:
Adds a render function with support for source maps.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,0 +1,5 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+export {render} from './src/node-sass/render';
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -2,4 +2,9 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-export {render} from './src/node-sass/render';
+export {
+  render,
+  RenderOptions,
+  RenderResult,
+  RenderError,
+} from './src/node-sass/render';
--- a/lib/src/node-sass/render.test.ts
+++ b/lib/src/node-sass/render.test.ts
@@ -2,9 +2,10 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-import * as p from 'path';
 import {promises as fs} from 'fs';
+import * as p from 'path';
 import {RawSourceMap} from 'source-map';
+import {pathToFileURL} from 'url';
 
 import * as sandbox from '../../../spec/helpers/sandbox';
 import {expectEqualIgnoringWhitespace} from '../../../spec/helpers/utils';
@@ -44,7 +45,7 @@ describe('render', () => {
     });
 
     it('enforces that one of data and file must be set', async done => {
-      render({}, error => {
+      render({file: ''}, error => {
         expect(error!.message).toBe(
           'Either options.data or options.file must be set.'
         );
@@ -145,7 +146,7 @@ describe('render', () => {
       await sandbox.run(async () => {
         await fs.writeFile(sassPath, 'a {b: c}');
         await render(
-          {data: `@import "${p.resolve(sassPath)}"`},
+          {data: `@import "${pathToFileURL(p.resolve(sassPath))}"`},
           (_, result) => {
             expectEqualIgnoringWhitespace(result!.css.toString(), 'a {b: c;}');
           }
@@ -262,9 +263,9 @@ describe('render', () => {
 
     describe('basic invocation', () => {
       const outFile = 'out.css';
-      let expectedMap;
-      let map;
-      let css;
+      let expectedMap: RawSourceMap;
+      let map: RawSourceMap;
+      let css: string;
 
       beforeAll(async done => {
         const data = 'a {b: c}';
--- a/lib/src/node-sass/render.test.ts
+++ b/lib/src/node-sass/render.test.ts
@@ -1,0 +1,722 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import * as p from 'path';
+import {promises as fs} from 'fs';
+import {RawSourceMap} from 'source-map';
+
+import * as sandbox from '../../../spec/helpers/sandbox';
+import {expectEqualIgnoringWhitespace} from '../../../spec/helpers/utils';
+import {compileString} from '../compile';
+import {render, RenderError, RenderResult} from './render';
+
+describe('render', () => {
+  const sassPathWithoutExtension = 'test';
+  const sassPath = `${sassPathWithoutExtension}.scss`;
+
+  describe('input', () => {
+    it('renders a file from an absolute path', async done => {
+      await sandbox.run(async () => {
+        await fs.writeFile(sassPath, 'a {b: c}');
+        await render({file: p.resolve(sassPath)}, (_, result) => {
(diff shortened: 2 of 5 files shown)


lib/src/node-sass/render.ts
```diff
@@ -0,0 +1,215 @@
...
+function newRenderResult(
+  options: RenderOptions,
+  start: number,
+  css: string,
+  sourceMap?: RawSourceMap
+): RenderResult {
+  const end = Date.now();
+  let sourceMapBytes;
+
+  if (wasSourceMapRequested(options)) {
```
nex3, line 127: Can you just check `if (sourceMap)` here? It's only going to be set if `options` requests it, right?

lib/src/node-sass/render.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    if (isEmpty(options.outFile)) {
+      if (isEmpty(options.file)) {
+        sourceMap.file = 'stdin.css';
+      } else {
+        const extension = p.extname(options.file!);
+        sourceMap.file =
+          extension === ''
+            ? `${options.file}.css`
+            : options.file!.replace(new RegExp(`${extension}$`), '.css');
```
nex3, line 145: It's probably safer to use `substring()` to slice of `extension.length` characters here rather than a regexp.

lib/src/node-sass/render.ts
```diff
@@ -0,0 +1,215 @@
...
+    sourceMap.sourceRoot = options.sourceMapRoot ?? '';
+
+    const sourceMapPath =
+      typeof options.sourceMap === 'string'
+        ? (options.sourceMap as string)
+        : options.outFile + '.map';
+    const sourceMapDir = p.dirname(sourceMapPath);
+
+    if (isEmpty(options.outFile)) {
+      if (isEmpty(options.file)) {
```
nex3, line 138: If you just write `if (options.outFile)` or `if (options.file)`, then you won't have to add explicit non-null assertions below. It'll trigger if the user passes in an empty string, but that's invalid for a path anyway so I don't think we need to worry about it too much.

lib/src/node-sass/render.ts
```diff
@@ -0,0 +1,215 @@
...
+// Transforms the compilation result into an object that mimics the Node Sass
+// API format.
+function newRenderResult(
+  options: RenderOptions,
+  start: number,
+  css: string,
+  sourceMap?: RawSourceMap
+): RenderResult {
+  const end = Date.now();
+  let sourceMapBytes;
```
nex3, line 125: Why not declare this where it's assigned?

lib/index.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+export {render} from './src/node-sass/render';
```
nex3, line 5: Should this also export the named types that are defined in `render.ts`?

lib/src/node-sass/render.test.ts
```diff
@@ -0,0 +1,722 @@
...
+import * as p from 'path';
+import {promises as fs} from 'fs';
+import {RawSourceMap} from 'source-map';
+
+import * as sandbox from '../../../spec/helpers/sandbox';
+import {expectEqualIgnoringWhitespace} from '../../../spec/helpers/utils';
+import {compileString} from '../compile';
+import {render, RenderError, RenderResult} from './render';
+
+describe('render', () => {
```
nex3, line 14: Not something we need to worry about right now, but I dream of a future where the JS API tests are shared between implementations.

lib/src/node-sass/render.test.ts
```diff
@@ -0,0 +1,722 @@
...
+        });
+      });
+      done();
+    });
+
+    it('supports absolute path imports', async done => {
+      await sandbox.run(async () => {
+        await fs.writeFile(sassPath, 'a {b: c}');
+        await render(
+          {data: `@import "${p.resolve(sassPath)}"`},
```
nex3, line 148: Convert this path into a URL before using it as an `@import`.

lib/src/node-sass/render.test.ts
```diff
@@ -0,0 +1,722 @@
...
+      if (result.map === undefined)
+        throw Error('Expected render result to contain a sourceMap.');
+      return JSON.parse(result.map.toString());
+    }
+
+    describe('basic invocation', () => {
+      const outFile = 'out.css';
+      let expectedMap;
+      let map;
+      let css;
```
nex3, line 267: Give these types.

nex3 (approved):
Very nice! A few cleanup suggestions, but this LGTM.

lib/src/node-sass/render.ts
Awjin: True, this double check is unnecessary.

lib/src/node-sass/render.ts
Awjin: It needs to be declared here because it's part of the return value

lib/index.ts
Awjin: Yes, good catch

lib/src/node-sass/render.test.ts
Awjin: Yep, I think that would be ideal! Are you envisioning an addition to `sass-spec` that handles testing all of the JS implementation?

lib/src/node-sass/render.ts
Awjin: Good point. I've simplify the `isEmpty` check to `!` in other parts of the code as well, since we don't want empty strings for either of these values.

lib/src/node-sass/render.ts
```diff
@@ -0,0 +1,215 @@
...
+  // sourceMapContents?: boolean;
+  sourceMapEmbed?: boolean;
+  sourceMapRoot?: string;
+  // TODO(awjin): https://github.com/sass/embedded-host-node/issues/13
+  // importer
+  // includePaths
+  // indentedSyntax
+  // functions
+  // outputStyle
+}
```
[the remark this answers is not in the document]
Awjin: Thanks. TIL that if we have—

```ts
type options = ({
  file: string
} | {
  file?: string;
  data: string;
}) & {
  omitSourceMapUrl?: boolean;
  ...
};
```

—`options.file` is available, but `options.data` is invalid. TS only permits the lowest common denominator of properties, which makes sense, but required some casting gymnastics to get the proper typings. PTAL: https://github.com/sass/embedded-host-node/pull/29/commits/7ecd7098ca6dc23ca502cc821f47b17a780a9441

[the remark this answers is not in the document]
Awjin: Thanks. TIL that if we have—

```ts
type options = ({
  file: string
} | {
  file?: string;
  data: string;
}) & {
  omitSourceMapUrl?: boolean;
  ...
};
```

—`options.file` is available, but `options.data` is invalid. TS only permits the lowest common denominator of properties, which makes sense, but required some casting gymnastics to get the proper typings. PTAL: https://github.com/sass/embedded-host-node/pull/29/commits/0ea0a466667bba3d445251dd72ce40f79ac57d48

lib/src/node-sass/render.ts
```diff
@@ -73,43 +83,46 @@ export interface RenderError extends Error {
...
+        sourceMap: getSourceMap,
+        url: stringRequest.file ? pathToFileURL(stringRequest.file) : 'stdin',
+      })
+    : compile({path: fileRequest.file, sourceMap: getSourceMap});
+
+  compileSass
+    .then(css =>
+      callback(undefined, newRenderResult(options, start, css, sourceMap))
+    )
+    .catch(error => callback(newRenderError(error)));
```
nex3, line 118: Rather than calling `.then()` followed by `.catch()` here, it's better to call `.then()` with both an `onFulfilled` and an `onRejected` callback. Otherwise, if `callback` throws an error synchronously in `then()`, that error will get passed back to `callback`!

nex3 (approved):

PR #31, opened 2021-02-20
Allow building Embedded Protocol from source.

Awjin:
Supports downloading the protocol release, building from master, or
building from a specific ref.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "google-protobuf": "^3.11.4",
     "immutable": "^4.0.0-rc.12",
     "rxjs": "^6.5.5",
+    "semver": "^5.7.1",
     "tar": "^6.0.5"
   },
   "devDependencies": {
@@ -34,6 +35,7 @@
     "@types/jasmine": "^3.5.11",
     "@types/node": "^12.18.2",
     "@types/node-fetch": "^2.5.7",
+    "@types/semver": "^7.3.4",
     "@types/shelljs": "^0.8.8",
     "@types/tar": "^4.0.3",
     "@types/yargs": "^15.0.12",
--- a/tool/embedded-binaries.ts
+++ b/tool/embedded-binaries.ts
@@ -3,6 +3,7 @@
 // https://opensource.org/licenses/MIT.
 
 import * as p from 'path';
+import {satisfies} from 'semver';
 import {promises as fs, existsSync} from 'fs';
 import fetch, {RequestInit} from 'node-fetch';
 import {extract as extractTar} from 'tar';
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "google-protobuf": "^3.11.4",
     "immutable": "^4.0.0-rc.12",
     "rxjs": "^6.5.5",
+    "semver": "^5.7.1",
```
nex3, line 29: Shouldn't this be a dev dependency?

tool/embedded-binaries.ts
```diff
@@ -70,162 +71,216 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
-    tag_name: string;
-    name: string;
+// Downloads release asset into `outPath`. Returns the version that was
+// downloaded.
+async function downloadRelease(options: {
+  repo: string;
+  outPath: string;
+  version?: string;
+  assetUrl: (release: ReleaseInfo) => string;
+  tags?: boolean;
```
nex3, line 108: Document these parameters.

tool/embedded-binaries.ts
```diff
@@ -70,162 +71,216 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
-    if (version) {
-      const response = await fetch(
-        'https://api.github.com/repos/sass/dart-sass-embedded/releases/tags/${version}',
-        fetchOptions
-      );
-      if (!response.ok) throw Error(response.statusText);
-      latestRelease = JSON.parse(await response.text());
+    const releases = JSON.parse(await response.text());
+    if (options.version) {
+      for (let i = 0; i < releases.length; i++) {
```
nex3, line 127: I'm guessing the earliest releases are at the beginning of the list? If so, add a comment to that effect.

tool/embedded-binaries.ts
```diff
@@ -70,162 +71,216 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
-    };
-    if (process.env.TRAVIS === 'true' && process.env.GITHUB_AUTH) {
-      fetchOptions['headers'] = {
-        Authorization:
-          'Basic ' +
-          Buffer.from(`sassbot:${process.env.GITHUB_AUTH}`).toString('base64'),
-      };
-    }
+    const response = await fetch(
+      `https://api.github.com/repos/sass/${options.repo}/${
```
nex3, line 118: Note that this action is [paginated](https://docs.github.com/en/rest/guides/traversing-with-pagination), so just fetching it once won't necessarily get you all the results. It's not super likely to matter in practice, but if you are trying to check against all the different versions it's probably better to be robust.

We had talked about just checking the single most recent release and verifying that it matches the expected constraint—is there a reason you didn't end up going that direction?

tool/embedded-binaries.ts
```diff
@@ -70,162 +71,216 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
+    await fs.rmdir(p.join(outPath, subdir), {recursive: true});
   } catch (_) {
     // If the path doesn't exist yet, that's fine.
   }
 }
 
 // Gets the current platform's operating system. Throws if the operating system
 // is not supported by Dart Sass Embedded.
-function getOs(): string {
+function getOs(): 'linux' | 'macos' | 'windows' {
```
nex3, line 283: Note: Since the OS is going to be constant for the entire run of the script, you could potentially make this (and arch/file extension) constants rather than functions.

tool/embedded-binaries.ts
```diff
@@ -70,162 +71,216 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
-    await fs.writeFile(archivePath, releaseAsset);
+    await cleanEmbeddedDir(options.outPath, options.repo);
+    const zippedAssetPath = `${options.outPath}/${
+      options.repo
+    }${getArchiveFileExtension()}`;
+    await fs.writeFile(zippedAssetPath, releaseAsset);
     if (getOs() === 'windows') {
-      await extractZip(archivePath, {dir: `${process.cwd()}/${outPath}`});
+      await extractZip(zippedAssetPath, {
+        dir: p.join(process.cwd()),
```
nex3, line 167: What's this `p.join()` doing?

tool/embedded-binaries.ts
```diff
@@ -70,162 +71,216 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
-      fetchOptions['headers'] = {
-        Authorization:
-          'Basic ' +
-          Buffer.from(`sassbot:${process.env.GITHUB_AUTH}`).toString('base64'),
-      };
-    }
+    const response = await fetch(
+      `https://api.github.com/repos/sass/${options.repo}/${
+        options.tags ? 'tags' : 'releases'
+      }`,
```
nex3, line 120:
```suggestion
      `https://api.github.com/repos/sass/${options.repo}/` +
      `${options.tags ? 'tags' : 'releases'}`,
```

Style nit: I think it's usually clearer to have a string split across multiple lines with `+` rather than having a multiline interpolation block.

tool/embedded-binaries.ts
```diff
@@ -70,162 +71,216 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
+
+  console.log('Building dart-sass-embedded executable.');
+  shell.exec('pub run grinder protobuf pkg-standalone-dev', {
+    cwd: repoPath,
+    silent: true,
+  });
+}
+
+// Ensures that `outPath` exists and the `subdir` does not.
+async function cleanEmbeddedDir(
```
nex3, line 269: Since this is no longer specific to `sass_embedded`, it should probably have a more generic name.

tool/embedded-binaries.ts
```diff
@@ -70,162 +71,216 @@ export async function getEmbeddedProtocol(outPath: string): Promise<void> {
...
+  outPath: string,
+  subdir: string
```
nex3, line 271: Consider just passing in one joined directory here and passing `p.dirname(path)` to `fs.mkdir()`.

nex3 (changes requested):

package.json
Awjin: This one needs to be a dependency because it'll be used when the npm package gets installed.

tool/embedded-binaries.ts
Awjin: Oops, thanks for the catch.

tool/embedded-binaries.ts
Awjin: Ah, that's right—I fixed this to just check the latest release.

tool/embedded-binaries.ts
```diff
@@ -2,59 +2,103 @@
...
+    break;
+  case 'darwin':
+    OS = 'macos';
+    break;
+  case 'win32':
+    OS = 'windows';
+    break;
+  default:
+    throw Error(`Platform ${process.platform} is not supported.`);
+}
```
nex3, line 30:
```suggestion
const OS: 'linux' | 'macos' | 'windows' = (() => {
  switch (process.platform) {
   case 'linux':
      return 'linux';
    case 'darwin':
      return 'macos';
    case 'win32':
      return 'windows';
    default:
      throw Error(`Platform ${process.platform} is not supported.`);
  }
})();
```

This will let you declare these as `const` (since they are logically constants) while also having complex initializer logic.

tool/embedded-binaries.ts
```diff
@@ -39,17 +77,22 @@ export async function getEmbeddedProtocol(
...
-      version: options.version,
-      assetUrl: releaseInfo =>
-        `https://github.com/sass/${repo}/archive` +
-        `/${releaseInfo.name.replace(' ', '-')}` +
-        getArchiveFileExtension(),
-      tags: true,
-    });
-    fs.rename(p.join(outPath, `${repo}-${version}`), p.join(outPath, repo));
+      options.version,
+      true
```
nex3, line 83: Nit: make this a required named arg, since otherwise it's impossible to guess from reading this what `true` means here.

tool/embedded-binaries.ts
```diff
@@ -98,89 +140,74 @@ export async function getDartSassEmbedded(
...
-      }
-    } else {
-      releaseInfo = releases[0];
-    }
-
-    if (!releaseInfo) {
-      throw Error(`Could not satisfy constraint ${options.version}.`);
+    const latestRelease = JSON.parse(await response.text())[0];
+    if (version && !satisfies(latestRelease.name, version)) {
+      throw Error(`Latest release is not compatible with ${version}.`);
```
nex3, line 164:
```suggestion
      throw Error(`Latest release ${latestRelease.name} is not compatible with ${version}.`);
```

nex3 (approved):

tool/embedded-binaries.ts
Awjin: Good call.

PR #33, opened 2021-02-25
Add a postinstall script.

Awjin:
This downloads the latest release of the compiler after an end-user `npm install`s
the host.

See https://github.com/sass/embedded-host-node/issues/37

### DIFF
--- a/.dev
+++ b/.dev
@@ -1,0 +1,14 @@
+# Problem
+
+The npm post-install hook provides no way to differentiate a developer who runs
+`npm install` to develop this package from an end-user who runs `npm install` to
+get the package from npm. Only end-users need to download the Embedded Compiler
+binary during post-install.
+
+# Solution
+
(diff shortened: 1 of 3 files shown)


.dev
```diff
@@ -0,0 +1,9 @@
+The npm postinstall hook provides no good way to differentiate a developer who
```
Awjin, line 1: Thoughts on this solution? Does it seem brittle?

nex3 (commented):
I'm concerned about using a hook script in a way that seems to be unintended by `npm`. I feel like it sets us up for confusion if other JS devs start working on this package. I think it might be cleaner to just require that people run these scripts by hand.

Awjin:
Discussed offline: there's no principled way to distinguish between a dev and end-user install, so it looks like we need to have the `.dev` file to distinguish them.

jathak (approved):

PR #40, opened 2021-04-13
Specify compiler version for end-user download.

Awjin:
This specifies the compiler version in `package.json`. Each release of the host will download that specified version to the end-user. In the future, we can continue to manually update the `package.json` entry, or lean on the synchronized version system and have the host download the compiler that matches its own version.

### DIFF
--- a/tool/utils.ts
+++ b/tool/utils.ts
@@ -74,7 +74,6 @@ export async function getEmbeddedProtocol(options: {
     const latestRelease = await getLatestReleaseInfo({
       repo,
       tag: true,
-      // TODO(awjin): versionConstraint
     });
     await downloadRelease({
       repo,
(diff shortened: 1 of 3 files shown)


tool/utils.ts
```diff
@@ -118,16 +119,17 @@ export async function getDartSassEmbedded(options: {
...
 
   if (options.release) {
-    const latestRelease = await getLatestReleaseInfo({
-      repo,
-      // TODO(awjin): versionConstraint
-    });
+    const release = options.version
+      ? {tag_name: options.version, name: `sass_embedded ${options.version}`}
+      : await getLatestReleaseInfo({
+          repo,
```
nex3, line 125:
```suggestion
          repo
```

Nit: this'll save a few vertical lines after formatting :smiley:

nex3 (approved):

tool/utils.ts
Awjin: I'd like that too but the formatter is enforcing this comma's existence (probably because these are function arguments?)

PR #45, opened 2021-04-17
Fix download of dart-sass-embedded on Windows

richard67:
Running `node ./download-compiler-for-end-user.js` during the build process fails on Windows, see https://github.com/joomla/joomla-cms/pull/33170#issuecomment-821800324 .

This pull request fixes that.


Awjin:
Good catch, thank you.

PR #46, opened 2021-04-29
Postinstall test

Awjin:
Verify that `npm postinstall` is working properly on all OSes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,8 @@ name: CI
 
 env:
   PROTOC_VERSION: 3.x
-  DEFAULT_NODE_VERSION: 15.x  # If changing this, also change jobs.tests.strategy.matrix.node_version
+  DEFAULT_NODE_VERSION: 15.x  # If changing this, also change jobs.tests.strategy.matrix.node-version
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +59,34 @@ jobs:
...
+        with:
+          version: ${{ env.PROTOC_VERSION }}
+          repo-token: "${{ github.token }}"
+      - uses: dart-lang/setup-dart@v1
+        with: {sdk: stable}
+      - run: dart --version
+      - run: npm install
+      - run: rm ./.dev
+      - run: ./node_modules/.bin/ts-node ./tool/prepare-release.ts
+      - run: npm run postinstall
```
nex3, line 88: Should we do a step here to verify that the install is actually functional somehow?

Awjin: It's probably worth it to do a validation here. Maybe a smoke test that runs the binary and compiles a simple stylesheet?

nex3: Yeah, that sounds good. Doesn't have to be complicated at all, just something that verifies that the library can do anything at all.

PR #49, opened 2021-05-26
Fix data: URI in source map sources

ntkme:
In some cases, `dart-sass-embedded` returns `data:` URI for `source` in the `sourceMap.sources`, which would break the current code.

This PR fixes it by check if `source.startsWith('file://')` as detection for `file:` uri only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.0-beta.3
+
+- Properly handle `data:` URIs in sourceMap sources.
+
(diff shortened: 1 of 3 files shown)


Awjin (approved):
Thanks for the fix! I'm working on fixing the broken CI, after which this can be merged.

ntkme:
Pushed a version fixing the complain from the linter.

Awjin:
Thanks. I still need to make changes to get these CI tests passing.

PR #51, opened 2021-08-03
Fix ci

Awjin:


lib/src/node-sass/render.test.ts
```diff
@@ -10,41 +10,73 @@ import {pathToFileURL} from 'url';
...
+import {render, RenderError, RenderOptions, RenderResult} from './render';
 
 describe('render', () => {
   const sassPathWithoutExtension = 'test';
   const sassPath = `${sassPathWithoutExtension}.scss`;
 
+  function expectRenderResult(
+    renderOptions: RenderOptions,
+    runExpectations: (result: RenderResult) => void
+  ) {
```
nex3, line 22: Please declare return types for all functions.

lib/src/node-sass/render.test.ts
```diff
@@ -10,41 +10,73 @@ import {pathToFileURL} from 'url';
...
-    it('renders a file from an absolute path', async done => {
-      await sandbox.run(async () => {
+    it('renders a file from an absolute path', () => {
+      return sandbox.run(async () => {
         await fs.writeFile(sassPath, 'a {b: c}');
-        await render({file: p.resolve(sassPath)}, (_, result) => {
-          expectEqualIgnoringWhitespace(result!.css.toString(), 'a {b: c;}');
+
+        return expectRenderResult({file: p.resolve(sassPath)}, result => {
+          expectEqualIgnoringWhitespace(result.css.toString(), 'a {b: c;}');
```
nex3, line 57: If you're already in an async function, imo it's generally better to use `await` for the final statement rather than `return`. It makes it clear that the value wrapped in the promise is irrelevant, and it makes it less likely that you'll accidentally break it when refactoring (for example, if you wrap the `return` in a `try`/`catch` it won't catch asynchronous errors but an `await` will).

nex3 (approved):

PR #53, opened 2021-08-05
Implement Value and sassNull

Awjin:
See: https://github.com/sass/sass/blob/main/proposal/new-function-and-values-api.d.ts. This proposal is still under review and likely to change, in which case this implementation will need updating.

### DIFF
--- a/lib/src/value/null.test.ts
+++ b/lib/src/value/null.test.ts
@@ -1,0 +1,25 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {sassNull} from './null';
+
+describe('Sass null', () => {
+  it('is falsey', () => {
+    expect(sassNull.isTruthy).toBe(false);
+  });
+
+  it('is equal to itself', () => {
+    expect(sassNull.equals(sassNull)).toBe(true);
+  });
+
+  it("isn't any type", () => {
+    expect(sassNull.assertBoolean).toThrow();
+    expect(sassNull.assertColor).toThrow();
+    expect(sassNull.assertFunction).toThrow();
+    expect(sassNull.assertMap).toThrow();
+    expect(sassNull.tryMap()).toBe(null);
+    expect(sassNull.assertNumber).toThrow();
+    expect(sassNull.assertString).toThrow();
+  });
+});
--- a/lib/src/value/null.test.ts
+++ b/lib/src/value/null.test.ts
@@ -2,24 +2,32 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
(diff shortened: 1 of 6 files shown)


lib/src/value/null.test.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {sassNull} from './null';
+
+describe('Sass null', () => {
```
nex3, line 7: Also test `.realNull`.

lib/src/value/null.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  it('is falsey', () => {
+    expect(sassNull.isTruthy).toBe(false);
+  });
+
+  it('is equal to itself', () => {
```
nex3, line 12: Any equality tests should probably verify the hash code as well.

lib/src/value/value.ts
```diff
@@ -0,0 +1,151 @@
...
+  get isTruthy(): boolean {
+    return true;
+  }
+
+  /** Returns JS null if `this` is `sassNull`. Otherwise, returns `this`. */
+  get realNull(): Value | null {
+    return this;
+  }
+
+  private _asList?: List<Value>;
```
nex3, line 35: I'm not sure this is worth caching.

lib/src/value/value.ts
```diff
@@ -0,0 +1,151 @@
...
+   * from the end of the list.
+   *
+   * `sassIndex` must be...
+   * - a number, and
+   * - an integer, and
+   * - a valid index into `asList`.
+   *
+   * Otherwise, this throws an error.
+   */
+  sassIndexToListIndex(sassIndex: Value): number {
```
nex3, line 66: Since this verifies details about `sassIndex`, it should take a `name` parameter. Can you update the spec?

lib/src/value/value.ts
```diff
@@ -0,0 +1,151 @@
...
+    throw Error('Not implemented yet');
+  }
+
+  /**
+   * Casts `this` to `SassBoolean`; throws if `this` isn't a boolean.
+   *
+   * If `this` came from a function argument, `name` is the argument name
+   * (without the `$`) and is used for error reporting.
+   */
+  assertBoolean(name?: string): Value {
```
nex3, line 77: This should return a `SassBoolean`. Same with other `assert()` functions.

nex3 (changes requested):

lib/src/value/value.ts
Awjin: I reference this in the comment below—I'll narrow these return types in future PRs as I implement each value. For now, it helps to have these assert functions in place for testing.

lib/src/value/value.ts
Awjin: Yep, PR here: https://github.com/sass/sass/pull/3111

nex3 (approved):

PR #56, opened 2021-08-19
Implement SassBoolean.

Awjin:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sass-embedded",
-  "version": "1.0.0-beta.3",
+  "version": "1.0.0-beta.4-dev",
```
nex3, line 3: This isn't technically a valid semantic version... is npm okay with that? Should it be `1.0.0-dev` instead for now?

lib/src/value/boolean.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Cannot be constructed; exists only as an interface and the exported
+ * singletons.
+ */
+export interface SassBoolean extends Value {
+  value: boolean;
+}
+
+const trueHash = hash(true);
+const falseHash = hash(true);
```
nex3, line 20:
```suggestion
const falseHash = hash(false);
```

lib/src/value/boolean.ts
```diff
@@ -0,0 +1,57 @@
...
+
+class SassBooleanInternal extends Value implements SassBoolean {
```
nex3, line 22: Can we call this just `SassBoolean`, like we do for `SassNull`?

nex3 (changes requested):
It looks like this doesn't yet export `sassTrue`, `sassFalse`, and `SassBoolean` (the interface) from the top-level `sass.ts` file. Can you add that as well?

lib/src/value/boolean.ts
Awjin: Unfortunately no; the interface and class names clash.

package.json
Awjin: npm doesn't see this until we push a git tag. But good point about the invalid semantics, I'll changed this to `1.0.0.beta.4` without `-dev`

nex3 (approved):

PR #57, opened 2021-08-20
Implement Sass Number

Awjin:


lib/src/value/number.ts
```diff
@@ -0,0 +1,718 @@
...
+  get hasUnits(): boolean {
+    return !(this.numeratorUnits.isEmpty() && this.denominatorUnits.isEmpty());
+  }
+
+  assertNumber(): SassNumber {
+    return this;
+  }
+
+  /**
+   * If `value` is an integer according to `isInt`, returns it as an integer.
```
nex3, line 248: "as an integer" is a bit ambiguous since JS doesn't have a proper integer type. Maybe "rounded to that integer"?

lib/src/value/number.ts
```diff
@@ -0,0 +1,718 @@
...
+// Returns the value of converting `number` to new units.
+//
+// The units may be specified as lists of units (`newNumeratorUnits` and
+// `newDenominatorUnits`), or by providng a SassNumber `other` that contains the
+// desired units.
+//
+// Throws an error if `number` is not compatible with the new units. Coercing a
+// unitful number to unitless (or vice-versa) throws an error unless
+// specifically enabled with `coerceUnitless`.
+function convertOrCoerce(
```
nex3, line 571: If you're always passing `this` as the first argument, why not just make this a private instance method?

nex3 (approved):

PR #61, opened 2021-09-15
Implement SassString.

Awjin:


lib/src/value/string.ts
```diff
@@ -0,0 +1,124 @@
...
+import {Value} from './value';
+import {valueError} from '../utils';
+
+/** A SassScript string. */
+export class SassString extends Value {
+  private readonly textInternal: string;
+  private readonly hasQuotesInternal: boolean;
+
+  /** Creates a string with `text`, optionally with quotes. */
+  constructor(text: string, options?: {/** @default true */ quotes: boolean}) {
```
nex3, line 16: I know you know this, but reminder to make this optional once https://github.com/sass/sass/pull/3151 lands.

nex3 (approved):

lib/src/value/string.ts
Awjin: Thanks for the reminder :)

PR #63, opened 2021-09-16
Implement SassColor.

Awjin:
https://github.com/sass/sass-spec/pull/1729

### DIFF
--- a/lib/src/value/color.ts
+++ b/lib/src/value/color.ts
@@ -1,0 +1,373 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {Value} from './value';
+import {fuzzyAssertInRange, fuzzyEquals, fuzzyRound} from './utils';
+import {hash} from 'immutable';
+
+interface RgbColor {
+  red: number;
+  green: number;
+  blue: number;
+  alpha: number;
+}
+
+interface HslColor {
+  hue: number;
+  saturation: number;
+  lightness: number;
+  alpha: number;
+}
+
+interface HwbColor {
+  hue: number;
+  whiteness: number;
+  blackness: number;
+  alpha: number;
+}
+
+/** A SassScript color. */
+export class SassColor extends Value {
+  private redInternal?: number;
+  private greenInternal?: number;
+  private blueInternal?: number;
+  private hueInternal?: number;
+  private saturationInternal?: number;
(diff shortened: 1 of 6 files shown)


lib/src/value/color.ts
```diff
@@ -0,0 +1,373 @@
...
+    }
+    return this.lightnessInternal!;
+  }
+
+  /** `this`'s whiteness value. */
+  get whiteness(): number {
+    // Because HWB is (currently) used much less frequently than HSL or RGB, we
+    // don't cache its values because we expect the memory overhead of doing so
+    // to outweigh the cost of recalculating it on access.
+    return (Math.min(Math.min(this.red, this.green), this.blue) / 255) * 100;
```
nex3, line 191:
```suggestion
    return Math.min(this.red, this.green, this.blue) / 255 * 100;
```

`Math.min()` takes any number of arguments. Same with `Math.max()` below.

lib/src/value/color.ts
```diff
@@ -0,0 +1,373 @@
...
+  // `red`, `green`, and `blue`.
+  //
+  // Algorithm from https://en.wikipedia.org/wiki/HSL_and_HSV#RGB_to_HSL_and_HSV
+  private rgbToHsl(): void {
+    const scaledRed = this.red / 255;
+    const scaledGreen = this.green / 255;
+    const scaledBlue = this.blue / 255;
+
+    const max = Math.max(Math.max(scaledRed, scaledGreen), scaledBlue);
+    const min = Math.min(Math.min(scaledRed, scaledGreen), scaledBlue);
```
nex3, line 331: Same here

nex3 (approved):

Awjin:
@nex3 I'm confused as to how type checks passed for Dart Sass, but are broken here. Shouldn't `ts-node` be catching these?

lib/src/value/color.ts
```diff
@@ -0,0 +1,361 @@
...
+    } else {
+      this.saturationInternal = (100 * delta) / (2 - max - min);
+    }
+  }
+
+  // Computes `this`'s red`, `green`, and `blue` channels based on `hue`,
+  // `saturation`, and `value`.
+  //
+  // Algorithm from the CSS3 spec: https://www.w3.org/TR/css3-color/#hsl-color.
+  private hslToRgb(): void {
```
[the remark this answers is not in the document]
nex3: Round-trips between the host and the compiler are very expensive, so we want to minimize them if at all possible. The color conversion algorithms are ultimately pretty simple, so I think it's worth asking wrappers to implement.

PR #65, opened 2021-09-22
Enable pinning protocol to a release version

Awjin:


tool/prepare-dev-environment.ts
```diff
@@ -12,32 +12,62 @@ const argv = yargs(process.argv.slice(2))
...
-      version: argv['protocol-version'],
-      path: argv['protocol-path'],
-    });
-    await getDartSassEmbedded({
-      outPath,
-      version: argv['compiler-version'],
-      path: argv['compiler-path'],
-    });
+
+    if (argv['protocol-version']) {
```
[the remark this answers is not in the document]
stof: yargs has `.conflicts()` to mark options as mutually exclusive

tool/utils.ts
```diff
@@ -56,146 +57,107 @@ interface ReleaseInfo {
...
+  outPath: string,
+  options?:
+    | {
+        version: string;
+      }
+    | {
+        ref: string | 'main';
+      }
+    | {
+        path: string;
```
stof, line 130: this is not the wrong type, as the `path` option can be passed alongside others.

Thus, passing an object with extra properties is supported, so this does not prevent passing both options either. So I suggest simplifying it:

```typescript
export async function getDartSassEmbedded(
  outPath: string,
  options?: {
    version?: string;
    ref?: string;
    path?: string;
  }
)
```

tool/utils.ts
```diff
@@ -56,146 +57,107 @@ interface ReleaseInfo {
...
-  if (!response.ok) {
-    throw Error(
-      `Failed to get version info for ${options.repo}: ${response.statusText}`
-    );
-  }
-  const latestRelease = JSON.parse(await response.text())[0];
-  const latestVersion = options.tag
-    ? latestRelease.name
-    : latestRelease.tag_name;
+  const repoPath = 'path' in options ? options.path : p.join(BUILD_PATH, repo);
```
stof, line 157: what is the use case for this, as fetching does not respect `options.path` ?

tool/utils.ts
stof: ah sorry, I misunderstood what `path` is.
But the implementation still accept reading `options.path` when it fetched the repo itself

tool/utils.ts
```diff
@@ -56,146 +57,107 @@ interface ReleaseInfo {
...
+ * By default, downloads the release version specified in package.json. Throws
+ * if an error occurs.
  */
-export async function getDartSassEmbedded(options: {
-  outPath: string;
-  version?: string;
-  path?: string;
-  release?: boolean;
-}): Promise<void> {
+export async function getDartSassEmbedded(
```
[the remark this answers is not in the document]
Awjin: Originally I tried unifying these, but there were enough small diffs in the ways that the protocol and compiler get built to make this difficult.

tool/utils.ts
Awjin: If `path` is provided, the compiler will be built from the source at that path. If `ref` is provided, the compiler will be built from the source at that Github ref.

tool/utils.ts
Awjin: I understand what you're saying here. Since this is in practice called through the CLI, I'm going to prioritize unblocking that use case first, then revisit this later.

PR #68, opened 2021-10-12
List and Map

Awjin:
See https://github.com/sass/sass-spec/pull/1739

### DIFF
--- a/lib/src/value/list.ts
+++ b/lib/src/value/list.ts
@@ -93,9 +93,9 @@ export class SassList extends Value {
 
   equals(other: Value): boolean {
     if (
-      other instanceof SassMap &&
+      (other instanceof SassList || other instanceof SassMap) &&
       this.contentsInternal.isEmpty() &&
-      other.contents.isEmpty()
+      other.asList.isEmpty()
     ) {
       return true;
     }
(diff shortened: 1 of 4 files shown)


lib/src/value/map.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.contentsInternal = contents;
+  }
+
+  /** Returns an empty map. */
+  static empty(): SassMap {
+    return new SassMap(OrderedMap<Value, Value>());
+  }
+
+  /** The separator for `this`'s contents as a list. */
+  get separator() {
```
nex3, line 26:
```suggestion
  get separator(): ListSeparator {
```

lib/src/value/list.ts
```diff
@@ -0,0 +1,131 @@
...
+    ) {
+      return false;
+    }
+
+    for (let i = 0; i < this.contentsInternal.size; i++) {
+      const thisValue = this.contentsInternal.get(i)!;
+      const otherValue = other.asList.get(i)!;
+      if (!thisValue.equals(otherValue)) return false;
+    }
+    return true;
```
nex3, line 107: Can you just make this `return this.contentsInternal.equals(other.asList)`?

nex3 (approved):

lib/src/value/map.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  hashCode(): number {
+    return this.contents.isEmpty()
+      ? SassList.empty().hashCode()
+      : this.contents.reduce(
+          (accumulator, value, key) =>
+            accumulator ^ value.hashCode() ^ key.hashCode(),
+          0
+        );
```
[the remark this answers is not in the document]
Awjin: Yep, exactly

PR #71, opened 2021-10-15
Implement indentedSyntax option

danieka:
As far as I understand it `indentedSyntax` is only applicable when compiling a string, and not when compiling a file.

### DIFF
--- a/lib/src/compile.ts
+++ b/lib/src/compile.ts
@@ -10,7 +10,10 @@ import {PacketTransformer} from './embedded-compiler/packet-transformer';
 import {MessageTransformer} from './embedded-protocol/message-transformer';
 import {Dispatcher} from './embedded-protocol/dispatcher';
 import {deprotifyException} from './embedded-protocol/utils';
-import {InboundMessage} from './vendor/embedded-protocol/embedded_sass_pb';
+import {
+  InboundMessage,
+  Syntax,
+} from './vendor/embedded-protocol/embedded_sass_pb';
 
 /**
  * Compiles a path and returns the resulting css. Throws a SassException if the
@@ -42,6 +45,7 @@ export async function compileString(options: {
   source: string;
   sourceMap?: (sourceMap: RawSourceMap) => void;
   url?: URL | string;
+  indentedSyntax?: boolean;
 }): Promise<string> {
   // TODO(awjin): Create logger, importer, function registries.
 
@@ -49,6 +53,7 @@ export async function compileString(options: {
     source: options.source,
     sourceMap: !!options.sourceMap,
     url: options.url instanceof URL ? options.url.toString() : options.url,
+    indentedSyntax: !!options.indentedSyntax,
   });
 
   const response = await compileRequest(request);
@@ -77,11 +82,13 @@ function newCompileStringRequest(options: {
   source: string;
   sourceMap: boolean;
   url?: string;
+  indentedSyntax: boolean;
 }): InboundMessage.CompileRequest {
   // TODO(awjin): Populate request with importer/function IDs.
 
   const input = new InboundMessage.CompileRequest.StringInput();
   input.setSource(options.source);
+  if (options.indentedSyntax) input.setSyntax(Syntax['INDENTED']);
   if (options.url) input.setUrl(options.url.toString());
 
   const request = new InboundMessage.CompileRequest();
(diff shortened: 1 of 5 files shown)


danieka:
@Awjin @nex3 I've rebased against latest `main` so all tests pass now. I'd be grateful if you could review the PR 😄

lib/src/compile.ts
```diff
@@ -42,13 +45,15 @@ export async function compileString(options: {
   source: string;
   sourceMap?: (sourceMap: RawSourceMap) => void;
   url?: URL | string;
+  indentedSyntax?: boolean;
```
Awjin, line 48: We will eventually need to support all of the syntaxes specced here: https://github.com/sass/sass/blob/main/spec/js-api/options.d.ts.

So instead of `indentedSyntax?: boolean`, this should be `syntax?: Syntax` where `Syntax` is a new type: `export type Syntax = 'scss' | 'indented' | 'css';`.

Please also update the tests in `compile.test.ts`.

lib/src/node-sass/render.test.ts
```diff
@@ -141,6 +141,19 @@ describe('render', () => {
...
+          data: 'a\n\tb: c',
+          indentedSyntax: true,
+        },
+        (_, result) => {
+          expectEqualIgnoringWhitespace(result!.css.toString(), 'a {b: c;}');
+          done();
+        }
+      );
+    });
   });
```
Awjin, line 157: There should be a few more tests here to cover the syntaxes:
- `.sass` file
- `.scss` file
- string with indented syntax (done)
- string with scss syntax

Awjin (changes requested):
Thanks for the contribution and sorry for the delay. I left a couple comments.

lib/src/node-sass/render.test.ts
danieka: Thank you for your comment! I just wanted to make sure I'm not misunderstanding you. Wouldn't the test `renders a file from a relative path` cover the `.scss` case, and wouldn't the test `renders a string` cover the "string with scss syntax". Or do you prefer creating additional, explicit, test cases?

lib/src/compile.ts
```diff
@@ -10,7 +10,12 @@ import {PacketTransformer} from './embedded-compiler/packet-transformer';
 import {MessageTransformer} from './embedded-protocol/message-transformer';
 import {Dispatcher} from './embedded-protocol/dispatcher';
 import {deprotifyException} from './embedded-protocol/utils';
-import {InboundMessage} from './vendor/embedded-protocol/embedded_sass_pb';
+import {
+  InboundMessage,
+  Syntax as SyntaxMap,
+} from './vendor/embedded-protocol/embedded_sass_pb';
```
Awjin, line 16: I would suggest `import * as proto` and using the namespaced `proto.Syntax` to avoid arbitrarily renaming this import.

lib/src/node-sass/render.test.ts
Awjin: Good point, I'm fine with not adding explicit cases.

lib/src/compile.ts
```diff
@@ -72,16 +79,24 @@ function newCompileRequest(options: {
   return request;
 }
 
+const mapSyntaxes = {
+  indented: 'INDENTED',
+  scss: 'SCSS',
+  css: 'CSS',
+} as const;
+
```
Awjin, line 87: I'm not sure this map is adding much. Recommend in-lining this as a 3-way `if` statement in line 99

Awjin (changes requested):
Looking good. Could you add an entry in the CHANGELOG under `1.0.0-beta.5` describing this change?

lib/src/compile.ts
```diff
@@ -77,14 +81,21 @@ function newCompileStringRequest(options: {
...
   // TODO(awjin): Populate request with importer/function IDs.
 
-  const input = new InboundMessage.CompileRequest.StringInput();
+  const input = new proto.InboundMessage.CompileRequest.StringInput();
   input.setSource(options.source);
+
+  if (options.syntax === 'scss') input.setSyntax(proto.Syntax['SCSS']);
+  else if (options.syntax === 'indented')
+    input.setSyntax(proto.Syntax['INDENTED']);
+  else if (options.syntax === 'css') input.setSyntax(proto.Syntax['CSS']);
```
Awjin, line 94: Style nit:

```suggestion
  if (options.syntax === 'scss') {
    input.setSyntax(proto.Syntax['SCSS']);
  } else if (options.syntax === 'indented') {
    input.setSyntax(proto.Syntax['INDENTED']);
  } else if (options.syntax === 'css') {
    input.setSyntax(proto.Syntax['CSS']);
  }
```

Awjin (approved):
I had one more style nit, otherwise this LGTM. Thanks @danieka!

PR #76, opened 2021-11-02
Add includedFiles to render output

danieka:
I don't know if this fits with your thoughts on how to return stuff from the compilation. I'm open to feedback and modifying the PR.

### DIFF
--- a/lib/src/compile.test.ts
+++ b/lib/src/compile.test.ts
@@ -14,9 +14,9 @@ describe('compile', () => {
   describe('success', () => {
     describe('input', () => {
       it('compiles an SCSS string by default', async () => {
-        expect(await compileString({source: 'a {b: c}'})).toBe(
-          'a {\n  b: c;\n}'
-        );
+        expect(await compileString({source: 'a {b: c}'})).toMatchObject({
+          css: 'a {\n  b: c;\n}',
+        });
       });
 
       // TODO(awjin): compiles an SCSS string explicitly
@@ -27,7 +27,7 @@ describe('compile', () => {
         const path = resolve('test.scss');
         await fs.writeFile(path, 'a {b: c}');
 
-        expect(await compile({path})).toBe('a {\n  b: c;\n}');
+        expect(await compile({path})).toMatchObject({css: 'a {\n  b: c;\n}'});
         await fs.unlink(path);
       });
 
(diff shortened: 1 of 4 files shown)


lib/src/compile.test.ts
```diff
@@ -14,9 +14,9 @@ describe('compile', () => {
   describe('success', () => {
     describe('input', () => {
       it('compiles an SCSS string by default', async () => {
-        expect(await compileString({source: 'a {b: c}'})).toBe(
-          'a {\n  b: c;\n}'
-        );
+        expect(await compileString({source: 'a {b: c}'})).toMatchObject({
```
Awjin, line 17: `toMatchObject` is too permissive here, let's use the stricter `toEqual` instead. (same for the other tests below)

```suggestion
        expect(await compileString({source: 'a {b: c}'})).toEqual({
```

lib/src/compile.ts
```diff
@@ -12,14 +12,19 @@ import {Dispatcher} from './embedded-protocol/dispatcher';
 import {deprotifyException} from './embedded-protocol/utils';
 import {InboundMessage} from './vendor/embedded-protocol/embedded_sass_pb';
 
+type CompileResult = {
```
Awjin, line 15:
```suggestion
export type CompileResult = {
```

Could you also export this from `lib/index.ts` so that it becomes part of the top-level API?

lib/src/compile.test.ts
```diff
@@ -27,24 +27,24 @@ describe('compile', () => {
...
       it('outputs in expanded mode by default', async () => {
-        expect(await compileString({source: 'a {b: c}'})).toBe(
-          'a {\n  b: c;\n}'
-        );
+        expect(await compileString({source: 'a {b: c}'})).toMatchObject({
+          css: 'a {\n  b: c;\n}',
+        });
       });
 
       // TODO(awjin): outputs in expanded mode explicitly
```
Awjin, line 50: Please also add a test for a compilation that has `includedUrls` in the result.

lib/src/node-sass/render.ts
```diff
@@ -183,6 +187,8 @@ function newRenderResult(
     }
   }
 
+  const includedFiles = loadedUrls.map(url => url.replace('file://', ''));
```
Awjin, line 190: This needs to handle file URLs properly on Windows, too.

```suggestion
  const includedFiles = loadedUrls.map(url => {
    return url.startsWith('file://') ? fileURLToPath(url) : url;
  });
```

Awjin (changes requested):

danieka:
I'm sorry I never got around to finishing this, as far as I can tell it has already been implemented since the I made this PR so I'm going to close it. Super excited about the progress you're making @nex3 @Awjin! Thank you 😄

PR #78, opened 2021-11-12
Support improvements to the new JS API

nex3:
See sass/sass#3200
See sass/sass-spec#1741


lib/src/value/color.ts
```diff
@@ -20,6 +20,13 @@ interface HslColor {
   alpha: number;
 }
 
+interface HwbColor {
+  hue: number;
+  whiteness: number;
+  blackness: number;
+  alpha: number;
```
stof, line 27: shouldn't the alpha channel be optional in those types ? The constructor has a check `color.alpha === undefined`

lib/src/value/color.ts
```diff
@@ -30,102 +37,77 @@ export class SassColor extends Value {
...
+        0,
+        100,
+        'lightness'
+      );
     } else {
-      this.hueInternal = color.hue;
-      this.saturationInternal = color.saturation;
-      this.lightnessInternal = color.lightness;
+      // From https://www.w3.org/TR/css-color-4/#hwb-to-rgb
+      const scaledHue = (color.hue % 360) / 360;
```
stof, line 78: be careful. In JS, `-3 % 360` gives `-3`, not `357` (the sign of the return value matches the sign of the left operand, not of the right one, unlike in dart)

lib/src/value/color.ts
```diff
@@ -344,6 +311,13 @@ export class SassColor extends Value {
   }
 }
 
+// A helper for converting HWB colors to RGB.
+function toRgb(hue: number, scaledWhiteness: number, scaledBlackness: number) {
```
stof, line 315: should probably be named `hwbToRgb` rather than just `toRgb` to make it clearer (we might have `hslToRgb`)

lib/src/value/list.ts
```diff
@@ -26,14 +32,18 @@ export class SassList extends Value {
    * `brackets`.
    */
   constructor(
```
stof, line 34: This should probably be defined using overloads, to reject `construct(ConstructorOptions, ConstructorOptions)` from valid calls in the type checker

lib/src/value/color.ts
```diff
@@ -30,102 +37,77 @@ export class SassColor extends Value {
   private lightnessInternal?: number;
   private readonly alphaInternal: number;
 
-  private constructor(color: RgbColor | HslColor) {
+  constructor(color: RgbColor | HslColor | HwbColor) {
```
Awjin, line 40: This can be defined as constructor overloads to enforce the type of each param.

```suggestion
  constructor(color: RgbColor);
  constructor(color: HslColor);
  constructor(color: HwbColor);
  constructor(color: RgbColor | HslColor | HwbColor) {
```

(the same goes for the rest of these overloaded constructors and functions)

Awjin (changes requested):

lib/src/value/color.ts
nex3: Oh interesting, I didn't realize that was different from a type system perspective.

lib/src/value/color.ts
nex3: Oh yikes, thanks for the heads up!

lib/src/value/color.ts
stof: using constructor overloads there is exactly the opposite of the typescript best practice.

And it does not provide any benefit. The first argument can indeed be either `RgbColor` or `HslColor` or `HwbColor` and that's exactly what the union type tells to the compiler.

Awjin: If this is just a union type, the compiler wouldn't complain about this:

```ts
new SassColor({
  red: 255,
  green: 255,
  blue: 255,
  hue: 0,
  saturation: 0,
  lightness: 100,
})
```

whereas overloads properly catch the error.

stof: Are you sure about that ? From the doc of typescript, the 2 cases should be equivalent

stof: I opened https://github.com/microsoft/TypeScript/issues/46873 to get the feedback from the TS team on that case.

stof: Note that this works only when using an object literal in the argument. As soon as the object comes from elsewhere (a variable being used), extra properties will be allowed (as that would be a subtype of the RgbColor type).
If you actually want to forbid passing `lightness` in a RgbColor (which might make sense in that context), you would have to define those HSL properties with type `never` in the RgbColor type.

lib/src/value/string.ts
```diff
@@ -13,10 +13,19 @@ export class SassString extends Value {
   private readonly hasQuotesInternal: boolean;
 
   /** Creates a string with `text`, optionally with quotes. */
-  constructor(text: string, options?: {/** @default true */ quotes?: boolean}) {
+  constructor(
+    textOrOptions?: string | {quotes?: boolean},
+    options?: {quotes?: boolean}
+  ) {
```
Awjin, line 19: Let's make this an overload like SassList
```suggestion
  constructor(text: string, options?: ConstructorOptions);
  constructor(options?: ConstructorOptions);
  constructor(
    textOrOptions?: string | ConstructorOptions,
    options?: ConstructorOptions
  ) {
```

lib/src/value/list.ts
```diff
@@ -28,12 +34,24 @@ export class SassList extends Value {
...
     contents: Value[] | List<Value>,
     options?: {
-      /** @default ',' */ separator?: ListSeparator;
-      /** @default false */ brackets?: boolean;
+      separator?: ListSeparator;
+      brackets?: boolean;
     }
```
Awjin, line 39: This can just be `options?: ConstructorOptions`

lib/src/value/list.ts
```diff
@@ -28,12 +34,24 @@ export class SassList extends Value {
...
+  );
+  constructor(options?: {separator?: ListSeparator; brackets?: boolean});
```
Awjin, line 41:
```suggestion
  constructor(options?: ConstructorOptions);
```

lib/src/value/color.ts
Awjin: I think the pass-through case is compelling enough to just make this a union type. We can add `never` fields in a follow-up PR to make the types mutually exclusive.

lib/src/value/color.ts
```diff
@@ -202,67 +196,55 @@ export class SassColor extends Value {
...
-  changeAlpha(alpha: number): SassColor {
-    return new SassColor({
-      red: this.red,
-      green: this.green,
-      blue: this.blue,
-      alpha: fuzzyAssertInRange(alpha, 0, 1, 'alpha'),
-    });
+  change(color: Partial<RgbColor>): SassColor;
+  change(color: Partial<HslColor>): SassColor;
+  change(color: Partial<HwbColor>): SassColor;
```
Awjin, line 203: In the same vein as the constructor:

```suggestion
```

Awjin (approved):

PR #79, opened 2021-11-14
Adds `includePaths` option to render()

denizsokullu:
Hi! I have been following this project closely and wanted to help out with a small contribution.


denizsokullu:
@nex3 Is there anything I can do to help this get reviewed? It's been sitting here for a while now.

lib/src/compile.test.ts
```diff
@@ -61,6 +61,36 @@ describe('compile', () => {
         expect(await compile({path})).toBe('a {\n  b: c;\n}');
         await fs.unlink(path);
       });
+
+      it('compiles imports using includePaths', async () => {
```
Awjin, line 65: Please also add a spec in `render.test.ts`.

lib/src/compile.ts
```diff
@@ -21,18 +21,21 @@ export type Syntax = 'scss' | 'indented' | 'css';
 export async function compile(options: {
   path: string;
   sourceMap?: (sourceMap: RawSourceMap) => void;
+  includePaths?: string[];
```
Awjin, line 24: Here and throughout: rename this to `loadPaths` to match the new JS API (https://github.com/sass/sass/blob/main/spec/js-api/options.d.ts#L74).

lib/src/compile.ts
```diff
@@ -66,13 +71,20 @@ export async function compileString(options: {
...
 }): proto.InboundMessage.CompileRequest {
   // TODO(awjin): Populate request with importer/function IDs.
 
   const request = new proto.InboundMessage.CompileRequest();
   request.setPath(options.path);
   request.setSourceMap(options.sourceMap);
 
+  for (const path of options.includePaths) {
+    const importer = new proto.InboundMessage.CompileRequest.Importer();
+    importer.setPath(path);
```
Awjin, line 84: Here and below: resolve `path` before passing it to the importer.

Awjin (changes requested):

lib/src/compile.ts
denizsokullu: I'm guessing you wanted me to Path.resolve the `path` string before it's passed into `importer.setPath(...)`.

lib/src/compile.test.ts
denizsokullu: Added. The render tests use a sandbox wrapper that the compile tests don't. Should I be using it in compile.test.ts? It seems to handle the folder/file cleanup.

nex3:
Thanks for the contribution!

PR #88, opened 2021-12-22
Use clone-linked-repo to check out the embedded protocol/compiler

nex3:
This defaults to checking out the versions listed in the pubspec, but
can be overridden for an individual pull request. Note that since the
main branch still always runs tests against specific released
versions, this requires that linked PRs be rolled into tagged releases
immediately after being landed.

The alternative would be to always run local tests against the main
branches of the protocol/compiler repos, but that means we may be
testing against a different version than we release against which
would be quite bad.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
       - run: npm install
       - name: npm run init
         run: |
-          npm run init \
+          npm run init -- \
             --protocol-path=embedded-protocol \
             --compiler-path=dart-sass-embedded \
             --api-path=language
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -57,8 +87,37 @@ jobs:
       - uses: dart-lang/setup-dart@v1
         with: {sdk: stable}
       - run: dart --version
+
+      - name: Read package.json
```
jathak, line 91: Is there a way to avoid the repetition here with a YAML alias, or is there something about GitHub Actions or this specific config that's preventing that?

jathak (approved):

.github/workflows/ci.yml
nex3: https://github.community/t/support-for-yaml-anchors/16128/41 :sob:

PR #89, opened 2021-12-22
Support the SassFunction type

nex3:
See sass/sass#2510
See sass/sass-spec#1752

### DIFF
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -15,9 +15,19 @@ import {PromiseOr, catchOr, thenOr} from './utils';
 import {Protofier} from './protofier';
 import {Value} from './value';
 
+/**
+ * The next ID to use for a function. The embedded protocol requires that
+ * function IDs be globally unique.
+ */
+let nextFunctionID = 0;
+
+/**
+ * Tracks functions that are defined on the host so that the compiler can
+ * execute them.
+ */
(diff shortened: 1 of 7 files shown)


lib/src/function-registry.ts
```diff
@@ -15,9 +15,19 @@ import {PromiseOr, catchOr, thenOr} from './utils';
...
+let nextFunctionID = 0;
+
+/**
+ * Tracks functions that are defined on the host so that the compiler can
+ * execute them.
+ */
 export class FunctionRegistry<sync extends 'sync' | 'async'> {
   private readonly functionsByName = new Map<string, CustomFunction<sync>>();
-  private readonly functionsById = new Array<CustomFunction<sync>>();
+  private readonly functionsById = new Map<number, CustomFunction<sync>>();
```
jathak, line 30: I'm not well-versed in TypeScript style. Is there a particular reason to use a Map with number keys instead of an Array here?

jathak (approved):

lib/src/function-registry.ts
nex3: Since functions are required to have unique IDs across all compilations, if this gets invoked multiple times during a compilation it may have IDs that are much larger than the beginning of the array. While JS does support sparse arrays, I think it's a lot clearer to represent this more explicitly using a map.

PR #91, opened 2021-12-23
Support custom importers

nex3:
Closes sass/sass#2509
See sass/sass-spec#1756
See sass/dart-sass-embedded#59

### DIFF
--- a/lib/src/compile.ts
+++ b/lib/src/compile.ts
@@ -2,7 +2,6 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-import * as p from 'path';
 import {Observable} from 'rxjs';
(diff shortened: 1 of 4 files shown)


lib/src/compile.ts
```diff
@@ -121,12 +125,6 @@ function newCompileRequest(
       throw new Error(`Unknown options.style: "${options?.style}"`);
   }
 
-  for (const path of options?.loadPaths ?? []) {
```
jathak, line 124: Is there a reason why this is deleted? Are importers created for load paths somewhere else?

jathak (approved):

lib/src/compile.ts
nex3: Yep, line 38 of `importer-registry.ts`

PR #96, opened 2022-01-12
Add basic legacy function infrastructure and boolean/null support

nex3:
See sass/sass-spec#1767

### DIFF
--- a/lib/src/value/boolean.ts
+++ b/lib/src/value/boolean.ts
@@ -19,9 +19,22 @@ export interface SassBoolean extends Value {
 const trueHash = hash(true);
 const falseHash = hash(false);
 
-class SassBooleanInternal extends Value implements SassBoolean {
+export class SassBooleanInternal extends Value implements SassBoolean {
+  // Whether callers are allowed to construct this class. This is set to
+  // `false` once the two constants are constructed so that the constructor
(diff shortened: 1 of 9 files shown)


lib/src/value/boolean.ts
```diff
@@ -48,10 +61,25 @@ class SassBooleanInternal extends Value implements SassBoolean {
   toString(): string {
     return this.value ? 'sassTrue' : 'sassFalse';
   }
+
+  // Legacy API support
+
+  static TRUE: SassBooleanInternal;
+  static FALSE: SassBooleanInternal;
```
Awjin, line 68: It's unfortunate that legacy fields are polluting this class. Is there any way these could exist solely in the wrapper layer?

nex3: Not if we want to re-use the non-legacy boolean constants to avoid an extra layer of wrapping. The upside is that this class isn't even visible through the modern API unless someone does something weird like `sassTrue.constructor`, so it's unlikely anyone will actually notice these.

Awjin (approved):

lib/src/value/boolean.ts
Awjin: > this class isn't even visible through the modern API

Ah, that's what I was missing. Sounds good

PR #98, opened 2022-01-13
Add support for the legacy Color value API

nex3:
See sass/sass-spec#1769

### DIFF
--- a/lib/src/legacy/value/base.ts
+++ b/lib/src/legacy/value/base.ts
@@ -11,7 +11,5 @@ import {Value} from '../../value';
 export class LegacyValueBase<T extends Value> {
   inner: T;
 
-  constructor(inner: T) {
-    this.inner = inner;
-  }
+  constructor(readonly inner: T) {}
 }
--- a/lib/src/legacy/value/base.ts
+++ b/lib/src/legacy/value/base.ts
@@ -9,5 +9,5 @@ import {Value} from '../../value';
  * detect legacy values and extract their inner value objects.
  */
 export class LegacyValueBase<T extends Value> {
-  constructor(readonly inner: T) {}
+  constructor(public inner: T) {}
 }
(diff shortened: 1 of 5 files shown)


lib/src/legacy/value/base.ts
```diff
@@ -0,0 +1,17 @@
...
+ * A base class for legacy value types. A shared base class makes it easier to
+ * detect legacy values and extract their inner value objects.
+ */
+export class LegacyValueBase<T extends Value> {
+  inner: T;
+
+  constructor(inner: T) {
+    this.inner = inner;
+  }
+}
```
Awjin, line 17:
```suggestion
  constructor(public inner: T) {}
}
```

lib/src/legacy/value/color.ts
```diff
@@ -0,0 +1,83 @@
...
+  getA(): number {
+    return this.inner.alpha;
+  }
+
+  setA(value: number): void {
+    this.inner = this.inner.change({alpha: clamp(value, 0, 1)});
+  }
+}
+
+Object.defineProperty(LegacyColor, 'name', {value: 'sass.types.Color'});
```
Awjin, line 78: What's the difference between this and a `static` member?

Awjin (approved):

lib/src/legacy/value/color.ts
nex3: A static member is semantically equivalent to assigning a field (in this case, `LegacyColor.name = 'sass.types.Color'`). But the `Function.name` field is read-only, so that breaks. Defining a property allows us to work around that and override it anyway.

lib/src/legacy/value/base.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import {Value} from '../../value';
+
+/**
+ * A base class for legacy value types. A shared base class makes it easier to
+ * detect legacy values and extract their inner value objects.
+ */
+export class LegacyValueBase<T extends Value> {
+  inner: T;
+
```
Awjin, line 13: This should be removed too.

```suggestion
```

nex3: Oops!

PR #100, opened 2022-01-14
Support the rest of the legacy JS function API

nex3:
See sass/sass-spec#1771

### DIFF
--- a/lib/src/legacy/value/index.ts
+++ b/lib/src/legacy/value/index.ts
@@ -17,3 +17,6 @@ export const Map = LegacyMap;
 export const Null = SassNull;
 export const Number = LegacyNumber;
 export const String = LegacyString;
+
+// For the `sass.types.Error` object, we just re-export the native Error class.
+export const Error = global.Error;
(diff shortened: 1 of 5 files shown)


lib/src/legacy/value/index.ts
```diff
@@ -17,3 +17,6 @@ export const Map = LegacyMap;
 export const Null = SassNull;
 export const Number = LegacyNumber;
 export const String = LegacyString;
+
+// For the `sass.types.Error` object, we just re-export the native Error class.
+export const Error = global.Error;
```
Awjin, line 22: I think this is missing some fields from the legacy error object.

nex3: No, this isn't the `LegacyError` class—this is the `sass.types.Error` class that's only used for async custom functions to communicate errors back to the caller. I'm not sure why the tests are failing, they pass locally for me.

nex3: Oh I see, I think we just need to update the compiler version.

Awjin (approved):

PR #102, opened 2022-01-19
Support the legacy JS importer API

nex3:
See sass/sass-spec#1773

### DIFF
--- a/lib/src/legacy/importer.ts
+++ b/lib/src/legacy/importer.ts
@@ -26,14 +26,6 @@ import {
 } from '../vendor/sass';
 
 /**
-
- * Special URL protocols we use to for the legacy importer. These are generally
- * not exposed to user code, but used internally to track
-
- */
-export const protocol = {};
-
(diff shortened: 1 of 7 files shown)


lib/src/legacy/importer.ts
```diff
@@ -0,0 +1,274 @@
...
+  LegacyImporterResult,
+  LegacyImporterThis,
+  LegacyPluginThis,
+  LegacySyncImporter,
+} from '../vendor/sass';
+
+/**
+
+ * Special URL protocols we use to for the legacy importer. These are generally
+ * not exposed to user code, but used internally to track
```
Awjin, line 31: Nit: incomplete?

Awjin (approved):

lib/src/legacy/importer.ts
nex3: Yeah, this entire variable is left over from an abortive refactor.

PR #107, opened 2022-01-28
Fix the linking of embedded-compiler PRs

stof:
Without those quotes, `--compiler-path=dart-sass-embedded` is not of the variable value, but a separate command being run, due to the space.


stof:
Note that this only fixes half of the issue. https://github.com/sass/clone-linked-repo/pull/6 is needed too.

nex3 (approved):

PR #108, opened 2022-01-28
Fix the `--no-compiler` flag for `npm init`

nex3:
Apparently, Yargs silently fails when you try to define a flag
beginning with `--no-`. This works around that by instead defining a
`--compiler` flag that default to `true`, at the cost of having
somewhat worse documentation and error messages.


Awjin (approved):
That is quite the unexpected behavior

Issue #112, opened 2022-02-02
Compiler caused error: Missing mandatory field Value.value


joneff:
First of all, bravo, for all the hard work! I am working with especially large sass project, so sass-embedded brings significant speed increase over dart-sass.

I am using a sort of hack to get all sass variables, along with their type and value like so:

```scss
// all.scss
$var1: 1;
$var2: 2px;

@import "PATH_TO_NODE_MODULES/bootstrap/scss/bootstrap.scss";
```

```scss
// variables.scss
@use "sass:meta";
@use "all.scss";

$_vars: meta.module-variables("all");

body {
    @each $key, $val in $_vars {
        var: resolve-var( unquote($key), meta.type-of( $val ), $val );
    }
}
```

Then I am compiling the file by passing a custom function like so:

```js
const fs = require('fs');
const path = require('path');
const sassEmbedded = require('sass-embedded');

const cwd = process.cwd();

let variablesJson = path.resolve( cwd, `dist/variables.json` );
let variablesScss = path.resolve( cwd, `dist/variables.scss` );
let content = {};

sassEmbedded.compile(variablesScss, {
    functions: {
        'resolve-var($key, $type, $value)': (args) => {
            const _key = args[0].toString();
            const _type = args[1].toString();
            const _val = args[2].toString();

            content[_key] = {
                type: _type,
                value: _val
            };

            return sassEmbedded.sassNull;
        }
    },
    logger: sassEmbedded.Logger.silent
});

fs.writeFileSync( variablesJson, JSON.stringify( content, null, 4 ) );
```

The above works with sass.

With sass embedded it throws an error:  `Compiler caused error: Missing mandatory field Value.value.` If I console log right after variable assignment, the last successful parsed variable is `$input-height-border`. So the one that's failing should be https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L839. It's worth noting that I've tried with calc and it also fails. other functions pass. 

Clarification, the line in question resorts to `calc`, so that's the one that fails.

On a side note, if I comment out bootstrap.scss and console.log(_val), I get different return types:
* dartSass yields `1` and `2px`
* dartEmbedded yields `1no unit` and `2px`

Awjin:
This looks like it might be a bug, but it is hard to tell from your code examples.

Could you provide a bare-minimum set of code that reproduces this bug?

joneff:
[...]

PR #115, opened 2022-02-21
Do not output `no units` when converting SassNumber to String

ntkme:
This PR solves one of the two issues reported in #112.

### DIFF
--- a/lib/src/value/number.ts
+++ b/lib/src/value/number.ts
@@ -613,10 +613,16 @@ export class SassNumber extends Value {
         }
       }
 
+      const unitSize = newNumerators.size + newDenominators.size;
       return valueError(
         `Expected $this to have ${
-          newNumerators.size + newDenominators.size > 1 ? 'units' : 'unit'
-        } ${unitString(newNumerators, newDenominators)}.`,
+          unitSize === 0
+            ? 'no units'
+            : `unit${unitSize > 1 ? 's' : ''} ${unitString(
+                newNumerators,
+                newDenominators
+              )}`
+        }.`,
         params.name
       );
     };
@@ -684,7 +690,7 @@ function unitString(
   denominators: List<string>
 ): string {
   if (numerators.isEmpty() && denominators.isEmpty()) {
-    return 'no units';
+    return '';
   }
 
   if (denominators.isEmpty()) {


ntkme:
The `unitString` function is used in two different places:

The first one should not add `no unit`:

https://github.com/sass/embedded-host-node/blob/db948177ec47e0da2a421de40c8d309d48e83606/lib/src/value/number.ts#L545-L550

The second one should add `no unit`:

https://github.com/sass/embedded-host-node/blob/db948177ec47e0da2a421de40c8d309d48e83606/lib/src/value/number.ts#L616-L622

This PR adds a boolean flag to control whether `no unit` is added or not.

ntkme:
@nex3 Something in the windows CI is flaky during `npm run init`. I tried to push the same commit three times and it failed twice, but worked the third time. Here is a failed run: https://github.com/sass/embedded-host-node/runs/6015818543?check_suite_focus=true

```
(node:5660) UnhandledPromiseRejectionWarning: Error: EPERM: operation not permitted, rename 'D:\a\embedded-host-node\embedded-host-node\build\embedded-protocol-1.0.0' -> 'D:\a\embedded-host-node\embedded-host-node\build\embedded-protocol'
```

Seems to be these lines:
https://github.com/sass/embedded-host-node/blob/464d9d5aeee83ab37da83bacfdb7dcec92440096/tool/utils.ts#L86-L89

ntkme:
@jathak Would you mind take a look when you get a chance?

jathak:
Yes, sorry for the delay

lib/src/value/number.ts
```diff
@@ -616,7 +616,7 @@ export class SassNumber extends Value {
       return valueError(
         `Expected $this to have ${
           newNumerators.size + newDenominators.size > 1 ? 'units' : 'unit'
-        } ${unitString(newNumerators, newDenominators)}.`,
+        } ${unitString(newNumerators, newDenominators, true)}.`,
```
jathak, line 619: Given this is the only place `unitString` is called with `verboseNoUnits = true`, I think it makes more sense to just eliminate the `verboseNoUnits` parameter and instead just only call `unitString` here if there are units, since right now this error's message would be `Expected $this to have unit no units`. It would be clearer to just error with `Expected $this to have no units` in that case.

jathak (commented):
Sorry for the delay on this!

ntkme:
Updated to address review feedback.

jathak (approved):

PR #123, opened 2022-04-01
Add support for npm installation (cross-platform) config

pomali:
This change enables to install package for different target platform/architecture.

For example on MacOS install Linux binary:
```
npm install  --arch=x64 --platform=linux sass-embedded
```
It's useful when you need to (cross-)build deployment artifact.

Inspired by sharp https://github.com/lovell/sharp/blob/d0c8e9564163cb8d1e1096c503f0c03d906f2666/lib/platform.js#L8


jathak:
Can you work on getting the tests passing and comment when you need the CI to be run again?

pomali:
I think it should pass tests now, after main got merged.

PR #124, opened 2022-04-03
Avoid resolving imports relative to working directory for compileString

ntkme:
This PR fixes https://github.com/sass/dart-sass-embedded/issues/82.

### DIFF
--- a/lib/src/compile.ts
+++ b/lib/src/compile.ts
@@ -2,6 +2,7 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+import * as p from 'path';
 import {Observable} from 'rxjs';
 import * as supportsColor from 'supports-color';
 
@@ -17,6 +18,7 @@ import {MessageTransformer} from './message-transformer';
 import {PacketTransformer} from './packet-transformer';
 import {SyncEmbeddedCompiler} from './sync-compiler';
 import {deprotofySourceSpan} from './deprotofy-span';
+import {legacyImporterProtocol} from './legacy/importer';
 
 export function compile(
   path: string,
@@ -87,10 +89,17 @@ function newCompileStringRequest(
   input.setSource(source);
   input.setSyntax(utils.protofySyntax(options?.syntax ?? 'scss'));
 
-  if (options?.url) input.setUrl(options.url.toString());
+  const url = options?.url?.toString();
+  if (url && url !== legacyImporterProtocol) {
+    input.setUrl(url);
+  }
 
   if (options && 'importer' in options && options.importer) {
     input.setImporter(importers.register(options.importer));
+  } else if (url === legacyImporterProtocol) {
+    const importer = new proto.InboundMessage.CompileRequest.Importer();
+    importer.setPath(p.resolve('.'));
+    input.setImporter(importer);
   }
 
   const request = newCompileRequest(importers, options);
--- a/lib/src/legacy/index.ts
+++ b/lib/src/legacy/index.ts
@@ -172,7 +172,9 @@ function convertStringOptions<sync extends 'sync' | 'async'>(
 
   return {
     ...modernOptions,
-    url: options.file ? pathToFileURL(options.file) : undefined,
+    url: options.file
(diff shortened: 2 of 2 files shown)


lib/src/compile.ts
```diff
@@ -89,8 +91,14 @@ function newCompileStringRequest(
 
   if (options?.url) input.setUrl(options.url.toString());
 
-  if (options && 'importer' in options && options.importer) {
-    input.setImporter(importers.register(options.importer));
+  const hasImporter = options && 'importer' in options && options.importer;
+  if (options?.url && !hasImporter) {
+    const importer = new proto.InboundMessage.CompileRequest.Importer();
+    importer.setPath(p.resolve('.'));
+    input.setImporter(importer);
```
ntkme, line 98: This makes sure the current directory in `FilesystemImporter('.')` is resolved on the host side instead of compiler side. It is not really necessary for the node host because as it would always restart the compiler so the compiler would have the same working directory as the host. However, as a reference implementation it is better to do the right thing here so that other implementations that re-use the compiler can follow the correct approach.

lib/src/compile.ts
```diff
@@ -89,8 +91,14 @@ function newCompileStringRequest(
...
+  } else {
+    const importer = hasImporter ? options.importer : noOpImporter;
```
ntkme, line 100: Explicitly send a no-op importer so that imports will not be resolved by the otherwise default `FilesystemImporter('.')` on the compiler side when both `url` and `importer` are `null`.

lib/src/legacy/index.ts
```diff
@@ -172,7 +172,9 @@ function convertStringOptions<sync extends 'sync' | 'async'>(
 
   return {
     ...modernOptions,
-    url: options.file ? pathToFileURL(options.file) : undefined,
+    url: options.file
+      ? pathToFileURL(options.file)
+      : new URL(legacyImporterProtocol),
```
ntkme, line 177: Legacy API needs to have a `FilesystemImporter('.')` as if it has an `url` even if it does not. Set a special value here to let the new API know that.

lib/src/compile.ts
```diff
@@ -87,10 +90,19 @@ function newCompileStringRequest(
   input.setSource(source);
   input.setSyntax(utils.protofySyntax(options?.syntax ?? 'scss'));
 
-  if (options?.url) input.setUrl(options.url.toString());
+  const url = options?.url?.toString();
+  if (url && url !== legacyImporterProtocol) {
```
ntkme, line 94: The special value `legacyImporterProtocol` means this is from Legacy API that even if it does not set a `url` it should behave as if it has one down below when setting the importer.

lib/src/compile.ts
```diff
@@ -87,9 +89,17 @@ function newCompileStringRequest(
...
 
-  if (options && 'importer' in options && options.importer) {
+  const hasImporter = options && 'importer' in options && options.importer;
+  if (url && !hasImporter) {
+    const importer = new proto.InboundMessage.CompileRequest.Importer();
+    importer.setPath(p.resolve('.'));
+    input.setImporter(importer);
+  } else if (hasImporter) {
     input.setImporter(importers.register(options.importer));
   }
```
ntkme, line 104: As of https://github.com/sass/dart-sass-embedded/pull/83, not setting importer would create a `noOp` importer on compiler side instead of `FilesystemImporter('.')`, thus it would fix https://github.com/sass/dart-sass-embedded/issues/82.

nex3:
Unless I'm mistaken, updating https://github.com/sass/dart-sass-embedded/pull/83 so that the compiler uses a filesystem importer in the right circumstances should mitigate the need for this PR. However, it would be good to also add a JS API spec to sass-spec that verifies that CWD-relative imports don't work with `compileString`.

ntkme:
@nex3 After reading that the base path in FileSystemImporter does not matter as it only will handle absolute path. I think you are right that this is not needed.

ntkme:
Looks like this can be simplified a bit by constructing proto importer inside the legacy compile.

ntkme:
Let me close this and open an new PR.

PR #127, opened 2022-04-08
Pass `--proto_path` to protoc

nex3:
This is necessary when passing a relative path outside the current
working directory to `--protocol-path`.

### DIFF
--- a/tool/utils.ts
+++ b/tool/utils.ts
@@ -3,7 +3,7 @@
 // https://opensource.org/licenses/MIT.
 
 import extractZip = require('extract-zip');
-import {promises as fs, existsSync} from 'fs';
+import {promises as fs, existsSync, mkdirSync} from 'fs';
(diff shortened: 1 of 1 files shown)


tool/utils.ts
```diff
@@ -262,6 +262,7 @@ function buildEmbeddedProtocol(repoPath: string): void {
       --plugin="protoc-gen-ts=${pluginPath}" \
       --js_out="import_style=commonjs,binary:." \
       --ts_out="." \
```
ntkme, line 264: The `js_out` and `ts_out` should probably match `proto_path`.

nex3: I think I actually want to make them explicitly `build/embedded-protocol`. Currently this is relying on the input path having that structure, which isn't a guarantee I want to enforce.

jathak (approved):

PR #134, opened 2022-04-20
Percent encode relative path like url.pathToFileURL

### LINKED ISSUE #133: sourcemap `sources` field gets url encoded
When a file is in a directory with a name including `@`, the `sources` field in sourcemap gets url encoded.
This does not happen with dart-sass.

```txt
.
└── src
    ├── @dir
    │   └── bar.scss
    └── foo.scss
```

```js
sass.render({
  file: path.resolve('src/foo.scss'),
  outFile: path.resolve('dist/foo.css'),
  sourceMap: true
})

sass.render({
  file: 'src/foo.scss',
  outFile: 'dist/foo.css',
  sourceMap: true
})
```

```text
[dart-sass] with resolve
{"version":3,"sourceRoot":"","sources":["../src/@dir/bar.scss","../src/foo.scss"],"names":[],"mappings":"AAAA;EACE;;;ACCF;EACE","file":"foo.css"}

-----------
[sass-embedded] with resolve
{"version":3,"sourceRoot":"","sources":["../src/%40dir/bar.scss","../src/foo.scss"],"names":[],"mappings":"AAAA;EACE;;;ACCF;EACE","file":"foo.css"}

-----------
[dart-sass] without resolve
{"version":3,"sourceRoot":"","sources":["file:///D:/documents/GitHub/sass-embedded-atmark-sourcemap-reproduction/src/@dir/bar.scss","file:///D:/documents/GitHub/sass-embedded-atmark-sourcemap-reproduction/src/foo.scss"],"names":[],"mappings":"AAAA;EACE;;;ACCF;EACE","file":"foo.css"}

-----------
[sass-embedded] without resolve
{"version":3,"sourceRoot":"","sources":["../src/%40dir/bar.scss","../src/foo.scss"],"names":[],"mappings":"AAAA;EACE;;;ACCF;EACE","file":"foo.css"}
```

### Step to reproduce
1. Clone https://github.com/sapphi-red/sass-embedded-atmark-sourcemap-reproduction.
2. Run `npm i`.
3. Run `node sass.js`
4. See output.

ntkme:
Fixes #133.

### DIFF
--- a/lib/src/utils.test.ts
+++ b/lib/src/utils.test.ts
@@ -15,5 +15,23 @@ describe('utils', () => {
         );
       }
     });
+
+    it('encode percent encoded string like `pathToFileURL`', () => {
+      const baseURL = pathToFileURL('').toString();
+      for (let i = 0; i < 128; i++) {
+        const lowercase = `%${i < 10 ? '0' : ''}${i.toString(16)}`;
+        expect(pathToUrlString(lowercase)).toEqual(
+          pathToFileURL(lowercase)
+            .toString()
+            .slice(baseURL.length + 1)
+        );
+        const uppercase = lowercase.toUpperCase();
+        expect(pathToUrlString(uppercase)).toEqual(
+          pathToFileURL(uppercase)
+            .toString()
+            .slice(baseURL.length + 1)
+        );
+      }
+    });
   });
 });
--- a/lib/src/utils.test.ts
+++ b/lib/src/utils.test.ts
@@ -1,0 +1,19 @@
+import {pathToFileURL} from 'url';
+import {pathToUrlString} from './utils';
+
+describe('utils', () => {
+  describe('pathToUrlString', () => {
+    it('encode relative path like `pathToFileURL`', () => {
+      const baseURL = pathToFileURL('').toString();
+      for (let i = 0; i < 128; i++) {
+        const char = String.fromCharCode(i);
+        const filename = `${i}-${char}`;
+        expect(pathToUrlString(filename)).toEqual(
+          pathToFileURL(filename)
+            .toString()
+            .slice(baseURL.length + 1)
+        );
+      }
+    });
+  });
+});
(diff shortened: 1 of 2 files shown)


lib/src/utils.ts
```diff
@@ -90,8 +90,14 @@ export function valueError(message: string, name?: string): Error {
...
 
-  const components = p.sep === '\\' ? path.split(/[/\\]/) : path.split('/');
-  return components.map(encodeURIComponent).join('/');
+  // percent encode relative path like `pathToFileURL`
+  return encodeURI(path)
+    .replace(/[#?]/g, encodeURIComponent)
+    .replace(
+      process.platform === 'win32' ? /%(5B|5C|5D|5E|7C)/g : /%(5B|5D|5E|7C)/g,
+      decodeURIComponent
+    )
```
nex3, line 99: I think this is supposed to be `encodeURIComponent`... currently this will convert the path `%5B` to the URL `[`, when it should be converted to `%255B`.

lib/src/utils.test.ts
```diff
@@ -0,0 +1,19 @@
+import {pathToFileURL} from 'url';
+import {pathToUrlString} from './utils';
+
+describe('utils', () => {
+  describe('pathToUrlString', () => {
+    it('encode relative path like `pathToFileURL`', () => {
```
nex3, line 6: Add a test for percent-encoding paths that are already percent-encoded, so we'll catch issues like the one below.

nex3 (changes requested):

lib/src/utils.ts
ntkme: This is intended. `pathname` of URL has a different [percent-encode set](https://url.spec.whatwg.org/#path-percent-encode-set) that some of the character does not need to be escaped.

ntkme: And it will not encode `%5B` to `[` because first `%` will be encoded as `%25` first, thus there is no `%5B` to be replaced.

lib/src/utils.test.ts
ntkme: Added.

nex3:
Ah, I understand now. Thanks!

PR #135, opened 2022-04-20
Make fetch happen

chrisuehlinger-techslice:
Resolves #117. I also ran into some race conditions occasionally when running `npm run init`, and it looks like there were some `fs.rename()` calls whose promises weren't being awaited, so I added `await`s for those. I checked that in as a separate commit in case you'd prefer I split this into two PRs.

PR #136, opened 2022-04-21
feat: support reading proxy from env

roland-reed:
Support reading proxy from environtment variables, as sometimes downloading from github.com fails.

Reading priority:
- HTTPS_PROXY
- HTTP_PROXY
- https_proxy
- http_proxy

Use `https-proxy-agent` under the hood.


ntkme:
I think #135 would be preferred over this as it is what npm uses.

jathak:
Going to close this in favor of that than.

PR #137, opened 2022-04-21
feat: Allow to change the Download Mirror, enabling corporate proxies… … and mirrors.

jweberde:
Allowing to set an installation variable in the package.json sass-embedded/downloadMirror, which allows to replace the default download location https://github.com/.

This is required in many corporate setups, where access to internet resources is restricted.


jathak:
Will using environment variables once #135 is merged satisfy this use case, or is there a separate need for this?

PR #149, opened 2022-05-31
Throw a better error when trying to install with musl-libc

nex3:


tool/utils.ts
```diff
@@ -155,6 +157,21 @@ export async function getDartSassEmbedded(
...
 
+/**
+ * Throws an informative error if we're running in a Linux environment that uses
+ * musl.
+ */
+async function checkForMusl(): Promise<void> {
+  if (process.platform !== 'linux') return;
+
+  const executable = await fs.readFile(process.execPath);
+  if (!executable.includes('libc.musl-x86_64.so.1')) return;
```
ntkme, line 168: We should also consider aarch64 as well as other arch here.

ntkme (changes requested):

tool/utils.ts
nex3: I'm not intimately familiar with what strings those are likely to leave in the executable. Can you give me a list of strings to search for?

ntkme: https://pkgs.alpinelinux.org/contents?file=&path=&name=musl&branch=edge&repo=main

ntkme:
@nex3 Sorry for going out of topic, but recently CI tests have been failing with either #148 or #150. It would be great if someone from the team can take a look and get those fixes merged. CI tests in dart-sass-embedded is also blocked by the same issues, as JS tests are added recently there.

tool/utils.ts
Goodwine: Would it make sense to check against `'/lib/ld-musl-'` and `'/lib/libc.musl-'` so it works regardless of architecture?

ntkme: It is extremely rare but multiple ld/libc can be installed in the same system at non-standard locations, so searching with `/lib/` might miss some rare cases. If we want good performance in searching, `libc.musl-` is probably the best.

Goodwine (approved):

PR #157, opened 2022-08-04
fix npm install

patricklx:
move yaml to dependencies so it will be installed, as its being used in the postinstall step.


MEthoD112 (approved):

nex3:
Thanks for the fix!

Issue #145, opened 2022-05-25
Install compiler executable using optional dependencies rather than postinstall


alan-agius4:
Currently, sass-embedded depends on a post-install script to download the native binary. In some cases corp environments  package managers are run with `ignore-script`. This causes the binary not be download. 

I propose that this download happens on first run instead of the mentioned hook. 

Example of the code needed
https://github.com/sass/embedded-host-node/blob/main/lib/src/compiler-path.ts
```ts
// Copyright 2021 Google LLC. Use of this source code is governed by an
// MIT-style license that can be found in the LICENSE file or at
// https://opensource.org/licenses/MIT.

import * as fs from 'fs';
import * as p from 'path';
import { spawnSync } from 'child_process';

/** The path to the embedded compiler executable. */
export const compilerPath = (() => {
  let sassCompilerPath = findCompilerPath();
  if (sassCompilerPath) return sassCompilerPath;

  // Try to download native compiler.
  spawnSync(process.execPath, [p.resolve(__dirname, '../../../download-compiler-for-end-user.js')]);

  sassCompilerPath = findCompilerPath();
  if (sassCompilerPath) return sassCompilerPath;

  throw new Error(
    "Embedded Dart Sass couldn't find the embedded compiler executable."
  );
})();

/** Finds the path to the embedded compiler executable. Returns `null` when it's not found. */
function findCompilerPath(): string | null {
  for (const path of ['vendor', '../../../lib/src/vendor']) {
    const executable = p.resolve(
      __dirname,
      path,
      `dart-sass-embedded/dart-sass-embedded${
        process.platform === 'win32' ? '.bat' : ''
      }`
    );

    if (fs.existsSync(executable)) return executable;
  }

  return null;
}
```

Other packages that depend on native componets such as esbuild, appear to be moving towards leveraging npm to pick the appropriate platform code via the `os` and `platform` package fields in the platform packages combined with `optionalDependencies`.  This has the advantage of leveraging npm's registry as well as package verification and lock file based versioning. 

Example: https://unpkg.com/browse/esbuild@0.14.39/package.json

nex3:
Doing something like esbuild does with optional dependencies seems valuable. I don't like the idea of running installations on first use, because that won't play well with people who are installing on one system and deploying on another or similar setups.

jathak:
Is there a reason why certain environments disable postinstall scripts?

nex3:
[...]

PR #158, opened 2022-08-12
Install compiler executable using optional dependencies

ahabhgk:
This PR resolves https://github.com/sass/embedded-host-node/issues/145.

I found the version is changed by `sassbot` when new sass releases, it might need the `sassbot` also changes the version of optionalDependencies and packages in the npm directory once this PR merged.


lib/src/compiler-path.ts
```diff
@@ -5,8 +5,35 @@
...
+    const executable = require.resolve(
+      `${
+        PACKAGES[`${platform} ${arch}`]
+      }/dart-sass-embedded/dart-sass-embedded${
+        platform === 'win32' ? '.bat' : ''
+      }`
+    );
+
+    if (fs.existsSync(executable)) return executable;
+  } catch (e) {
```
nex3, line 32: This should be typed:

```suggestion
  } catch (e: unknown) {
```

lib/src/compiler-path.ts
```diff
@@ -5,8 +5,35 @@
...
+    // ignore
```
nex3, line 33: Rather than ignoring all errors here, can we check the error type and only ignore errors that came from `require.resolve`?

lib/src/compiler-path.ts
```diff
@@ -5,8 +5,35 @@
...
+  try {
+    const executable = require.resolve(
+      `${
+        PACKAGES[`${platform} ${arch}`]
+      }/dart-sass-embedded/dart-sass-embedded${
+        platform === 'win32' ? '.bat' : ''
+      }`
+    );
+
+    if (fs.existsSync(executable)) return executable;
```
nex3, line 31: Is it possible for `require.resolve` to return a path that doesn't exist?

npm/linux-arm64/README.md
```diff
@@ -0,0 +1,3 @@
+# `@sass-embedded/linux-arm64`
+
+This is the **linux-arm64** binary for `sass-embedded`
```
nex3, line 3: These READMEs should link to the `sass-embedded` npm page.

npm/linux-arm64/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
nex3, line 1: Add `"description"` and `"author"` fields.

.github/workflows/ci.yml
```diff
@@ -198,6 +198,16 @@ jobs:
...
 
+      - name: Publish optional dependencies
+        run: |
+          cd npm/linux-arm64 && npm publish && cd ../..
+          cd npm/linux-ia32 && npm publish && cd ../..
+          cd npm/linux-x64 && npm publish && cd ../..
+          cd npm/macos-arm64 && npm publish && cd ../..
+          cd npm/macos-x64 && npm publish && cd ../..
+          cd npm/windows-ia32 && npm publish && cd ../..
+          cd npm/windows-x64 && npm publish && cd ../..
```
nex3, line 209: This will need the `NODE_AUTH_TOKEN` environment variable set, as in the action below.

lib/src/compiler-path.ts
```diff
@@ -5,8 +5,35 @@
...
+  'win32 ia32': '@sass-embedded/windows-ia32',
+  'win32 x64': '@sass-embedded/windows-x64',
+};
+
 /** The path to the embedded compiler executable. */
 export const compilerPath = (() => {
+  try {
+    const executable = require.resolve(
+      `${
+        PACKAGES[`${platform} ${arch}`]
```
nex3, line 25: Nit: I think referring directly to `process.platform` and `process.arch` here would be cleaner.

.github/workflows/ci.yml
nex3, line 209: Rather than putting a bunch of publishing logic in the individual `package.json` files, I think it would be better to run it here even if it makes these invocations more complex.

lib/src/compiler-path.ts
```diff
@@ -5,8 +5,35 @@
...
+const {platform, arch} = process;
+
+const PACKAGES: {[key: string]: string} = {
+  'linux arm64': '@sass-embedded/linux-arm64',
+  'linux ia32': '@sass-embedded/linux-ia32',
+  'linux x64': '@sass-embedded/linux-x64',
+  'darwin arm64': '@sass-embedded/macos-arm64',
+  'darwin x64': '@sass-embedded/macos-x64',
+  'win32 ia32': '@sass-embedded/windows-ia32',
+  'win32 x64': '@sass-embedded/windows-x64',
```
nex3, line 17: It's probably simpler to name the platform packages after the Node platform names, since they're going to be participating in the Node ecosystem. That will also allow you to get rid of this map entirely.

lib/src/compiler-path.ts
```diff
@@ -5,8 +5,35 @@
...
+};
+
 /** The path to the embedded compiler executable. */
 export const compilerPath = (() => {
```
nex3, line 21: Please update the error message here to explain that the user is missing one of the optional dependencies, and how to get that dependency.

tool/prepare-release.ts
```diff
@@ -50,6 +50,14 @@ async function sanityCheckBeforeRelease() {
     );
   }
 
+  for (const [dep, version] of Object.entries(pkg.optionalDependencies)) {
+    if (version !== releaseVersion) {
+      throw Error(
+        `optional dependency ${dep}'s version has not been updated to ${releaseVersion}.`
```
nex3, line 56:
```suggestion
        `optional dependency ${dep}'s version doesn't match ${releaseVersion}.`
```

I think it's clearer to just describe the problem here rather than try to explain its etiology.

nex3 (changes requested):
Thank you for taking this task on!

lib/src/compiler-path.ts
ahabhgk: like `@sass-embedded/win32-ia32` for windows and `@sass-embedded/darwin-x64` for macos?

ahabhgk:
Thanks for your review! Just did some updates according to the comments.

.github/workflows/ci.yml
```diff
@@ -198,6 +198,18 @@ jobs:
       - name: "Check we're not using a -dev version of the embedded compiler"
         run: jq -r '.["compiler-version"]' package.json | grep -qv -- '-dev$'
 
+      - name: Publish optional dependencies
+        env:
+          NODE_AUTH_TOKEN: '${{ secrets.NPM_TOKEN }}'
+        run: |
+          cd npm/linux-arm64 && npm_config_platform=linux npm_config_arch=arm64 npx ts-node ../../tool/prepare-optional-release.ts && npm publish && cd ../..
```
ahabhgk, line 205: Not sure if this is ok, since it manually sets the `npm_config_*` environment variables.

nex3: Can you pass these in as CLI parameters to `prepare-optional-release.ts` instead?

ahabhgk: Just changed it, please have a review.

npm/linux-arm64/package.json
ahabhgk: Have added it.

.github/workflows/ci.yml
ahabhgk: Tried CLI parameters, check https://github.com/sass/embedded-host-node/pull/158/commits/e6c9ea2557a7e341eaa5559689dcea2ca7e25211, but I think add a `prepare-optional` npm scripts is more simple and clean: `npm run prepare-optional --platform=linux --arch=arm64 && npm publish ./npm/linux-arm64`.

tool/prepare-optional-release.ts
```diff
@@ -0,0 +1,11 @@
+import {getDartSassEmbedded} from './utils';
+
+(async () => {
+  try {
+    const dirName = `${process.env.npm_config_platform}-${process.env.npm_config_arch}`;
```
nex3, line 5: It looks like this still isn't set up to take CLI parameters.

nex3 (changes requested):

tool/prepare-optional-release.ts
ahabhgk: Have tried CLI parameters in https://github.com/sass/embedded-host-node/commit/e6c9ea2557a7e341eaa5559689dcea2ca7e25211, which needs to change the way of getting `OS` and `ARCH` from a const variable to a function like `getOS(platform)` and `getArch(cpu)` since the OS and platform are different (OS: linux, macos, windows; platform: linux, darwin, win32), now is using an npm script `prepare-optional` and setup `npm_config_platform` and `npm_config_arch` by `npm run prepare-optional --platform=linux --arch=arm64`, I think it is more simple and clean, and it won't have too much invasive on existing implementations.
But if you insist the CLI parameters way is better, I can change it back to https://github.com/sass/embedded-host-node/commit/e6c9ea2557a7e341eaa5559689dcea2ca7e25211.

nex3: I think it's to be expected that this will change the logic of downloading packages, since the old code strongly assumed that it would always download packages for the current OS. I'd rather have it *always* take parameters and have `prepare-dev-environment` pass in the parameters for the current OS, actually.

ahabhgk: Clear enough, changed to CLI parameters way in all `prepare-*` scripts, please review again.

tool/utils.ts
```diff
@@ -16,10 +16,15 @@ const fetch = initMakeFetchHappen();
 
 shell.config.fatal = true;
 
-// The current platform's operating system. Throws if the operating system
+// Directory that holds source files.
+const BUILD_PATH = 'build';
+
+export type OSKind = 'linux' | 'macos' | 'windows';
```
nex3, line 22: It's probably a good idea to specify that this is specifically *Dart's* notion of OS kind, that it's different than Node's, and that you can convert from Node's to Dart's with `getOS()`. You could even name this something like `DartPlatform` to make it clearer.

Similarly for `ArchKind` below.

tool/utils.ts
```diff
@@ -16,10 +16,15 @@ const fetch = initMakeFetchHappen();
...
+export type ArchKind = 'ia32' | 'x64' | 'arm64';
+
+// Get the platform's operating system. Throws if the operating system
 // is not supported by Dart Sass Embedded.
-const OS: 'linux' | 'macos' | 'windows' = (() => {
-  const platform = process.env.npm_config_platform || process.platform;
+export function getOS(platform: string): OSKind {
```
nex3, line 27: Consider renaming this to something like `nodePlatformToDartPlatform` to clarify that it translates between the two.

Similarly for `getArch()` below.

tool/utils.ts
```diff
@@ -208,6 +218,7 @@ async function checkForMusl(): Promise<void> {
  */
 export async function getJSApi(
   outPath: string,
+  os: OSKind,
   options?: {ref: string} | {path: string}
 ): Promise<void> {
```
nex3, line 223: You shouldn't need to pass in an `os` parameter here. This only ever needs to know the *local* OS because it's just figuring out how best to create a local symlink, so it can just refer directly to `process.platform`.

Same for `buildEmbeddedProtocol()`.

nex3 (changes requested):

nex3 (approved):
Thanks again for working on this!

PR #161, opened 2022-08-31
Add support for 32-bit ARM executables on Linux

nex3:


npm/linux-arm/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "dart-sass-embedded/**/*"
+  ],
+  "engines": {
+    "node": ">=14.0.0"
+  },
+  "os": [
+    "linux"
+  ],
+  "cpu": [
+    "arm64"
```
jathak, line 18: Is this supposed to be `arm`?

nex3: It sure is! Good catch

jathak (approved):

PR #164, opened 2022-09-01
Fix error message for optional dependency

ntkme:
It should be a template string instead of a normal string to print the correct message.


nex3 (approved):

nex3:
Thanks!

PR #165, opened 2022-09-06
fix: remove unused dependencies and move dev dependencies

alan-agius4:
This change removes unused dependencies `semver`. And moves development only dependencies `extract-zip`, `shelljs`, `tar`,  `yaml` and `make-fetch-happen` to `devDependencies`

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,7 @@
   "main": "dist/lib/index.js",
   "types": "dist/types/index.d.ts",
   "files": [
-    "dist/**/*",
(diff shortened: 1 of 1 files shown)


nex3 (approved):

nex3:
Thanks!

nex3 (changes requested):
Looks like `google-protobuf` is still required at runtime.

alan-agius4:
Oops. re-added.

PR #167, opened 2022-11-09
Verify that importer 'contents' are a string-type

Goodwine:
Implements verification added on https://github.com/sass/dart-sass/pull/1816

See https://github.com/sass/sass-spec/pull/1838


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.55.1
```
ntkme, line 1: My understanding is that the CHANGELOG need to be added to main dart-sass repo and it will be synced to here by CI when we cut a new release.

nex3 (approved):

PR #168, opened 2022-12-09
Fix function name parsing with whitespace

ntkme:
This PR fixes function name parsing when whitespace exists between function name and `(`.

The following code works on sass but not sass-embedded due to function name being incorrectly parsed as `x ` (with extra whitespace at the end):

```
const sass = require('sass-embedded')

const result = sass.compileString('a{b:x()}', {
  functions: {
    'x ()': function () { return sass.sassTrue }
  }
})

console.log(result)
```

https://github.com/sass/sass-spec/pull/1854

### DIFF
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -37,7 +37,7 @@ export class FunctionRegistry<sync extends 'sync' | 'async'> {
         throw new Error(`options.functions: "${signature}" is missing "("`);
       }
 
-      this.functionsByName.set(signature.substring(0, openParen).trim(), fn);
+      this.functionsByName.set(signature.substring(0, openParen).trimEnd(), fn);
     }
   }
 
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -57,11 +57,11 @@ export class FunctionRegistry<sync extends 'sync' | 'async'> {
   call(
     request: OutboundMessage.FunctionCallRequest
   ): PromiseOr<InboundMessage.FunctionCallResponse, sync> {
-    const protofier = new Protofier(this);
-    const fn = this.get(request);
-
     return catchOr(
       () => {
+        const protofier = new Protofier(this);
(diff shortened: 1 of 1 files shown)


ntkme:
@nex3 Any comments?

nex3:
Sorry for the delay, we've got a lot of irons in the fire at the moment.

This seems like something that should be an error, since this isn't a valid function signature. [The spec](https://github.com/sass/sass/blob/2c655fcf43184746d952960de4921a1aa30771e9/spec/js-api/options.d.ts#L59-L86) is unfortunately not very clear about this, so we should update that. Unfortunately, since Dart Sass supports this we'll need to do a deprecation process.

The upshot here is that this should probably produce a deprecation warning, but we can do that in a separate PR once the language update goes through.

lib/src/function-registry.ts
```diff
@@ -57,11 +57,11 @@ export class FunctionRegistry<sync extends 'sync' | 'async'> {
   call(
     request: OutboundMessage.FunctionCallRequest
   ): PromiseOr<InboundMessage.FunctionCallResponse, sync> {
-    const protofier = new Protofier(this);
-    const fn = this.get(request);
-
     return catchOr(
       () => {
+        const protofier = new Protofier(this);
+        const fn = this.get(request);
```
nex3, line 63: Why this change?

ntkme: Because the error of this.get is not catches and raised back to compiler, the compilation will get stuck.

ntkme:
Most programming language allows a whitespace between function name and `(` so I think this is actually acceptable.

ntkme:
Sounds good. I guess since we are going to do deprecation, we should still want to make it work in sass-embedded like dart-sass for now, in a form that the complication does not get stuck, and then have both implementations behave the same on deprecation warnings.

nex3:
I'm just going to make it a breaking change—since the spec already says it should behave this way, I don't feel too bad about it.

PR #197, opened 2023-01-31
Compile protobufs using a modern protocol buffer compiler

nex3:


nex3:
Closing this out in favor of #198.

PR #201, opened 2023-02-09
Use buf to compile protobufs

nex3:
We were using @protobuf-ts/protoc, but that downloads protoc versions
live at runtime which causes many problems for CI. buf supports both a
standalone GitHub action and an npm installation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "minipass": "4.0.1",
     "node-fetch": "^3.3.0",
     "npm-run-all": "^4.1.5",
+    "path-equal": "^1.2.5",
     "shelljs": "^0.8.4",
     "source-map-js": "^1.0.2",
     "tar": "^6.0.5",
--- a/package.json
+++ b/package.json
@@ -38,14 +38,14 @@
   },
   "dependencies": {
     "@bufbuild/protobuf": "^1.0.0",
-    "@bufbuild/protoc-gen-es": "^1.0.0",
     "buffer-builder": "^0.2.0",
     "immutable": "^4.0.0",
     "rxjs": "^7.4.0",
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -38,14 +38,14 @@
...
     "@bufbuild/protobuf": "^1.0.0",
-    "@bufbuild/protoc-gen-es": "^1.0.0",
     "buffer-builder": "^0.2.0",
     "immutable": "^4.0.0",
     "rxjs": "^7.4.0",
     "supports-color": "^8.1.1"
   },
   "devDependencies": {
-    "@protobuf-ts/protoc": "^2.8.2",
+    "@bufbuild/buf": "^1.13.1-4",
```
Goodwine, line 47: does this need to be installed in ci.yml?

tool/get-embedded-protocol.ts
```diff
@@ -43,21 +43,9 @@ export async function getEmbeddedProtocol(
 
 // Builds the embedded proto at `repoPath` into a pbjs with TS declaration file.
 function buildEmbeddedProtocol(repoPath: string): void {
-  const proto = p.join(repoPath, 'embedded_sass.proto');
   const version = shell
-    .exec('npx protoc --version', {silent: true})
+    .exec('npx buf --version', {silent: true})
```
Goodwine, line 47: There are a few linter warnings in this file

Goodwine (changes requested):

package.json
nex3: I think once https://github.com/sass/dart-sass-embedded/pull/146 lands it'll pick it up from Node's internal PATH, but we'll see.

Goodwine (approved):

PR #203, opened 2023-02-16
Replace node-fetch with fetch

### LINKED ISSUE #202: Deploy npm failed
With the 1.58.0 release the last run of the `Deploy npm` job failed: https://github.com/sass/embedded-host-node/actions/runs/4069867897/jobs/7010247909.

The latest available version of sass-embedded on npm is still 1.57.1 (https://www.npmjs.com/package/sass-embedded). 

Is it possible to make 1.58.0 available on npm?

ntkme:
Latest node-fetch is ESM only, but our tsconfig still compile code to CommonJS, which breaks the import.

Since our release CI now runs on node 18 and node 18 has a built-in `fetch`, use it instead of `node-fetch`.

- Fixes #202


package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "sass-embedded",
-  "version": "1.58.0",
-  "protocol-version": "1.1.0",
-  "compiler-version": "1.58.0",
+  "version": "1.58.1",
+  "protocol-version": "1.2.0",
+  "compiler-version": "1.58.1",
```
nex3, line 5: Let's wait until Dart Sass 1.58.1 is released for these.

nex3 (changes requested):

package.json
ntkme: It's failing the test expecting these to be 1.58.1 as dart-sass-embedded from main branch is already on 1.58.1.

ntkme: dart-sass-embedded 1.58.1 release was partially broken as linux-arm64 and linux-arm were failing. So we would have to go 1.58.2 directly as next release.

nex3 (approved):

PR #204, opened 2023-02-16
Roll node-fetch back to a pre-ESM version

nex3:
See https://github.com/sass/embedded-host-node/issues/202


jathak (approved):

ntkme:
@nex3 @jathak Shall we consider #203 instead of this?

nex3:
Yeah, I missed that, that's a better option.

PR #212, opened 2023-03-21
Fix resolving against the working directory in legacy API

ntkme:
The issue here is that currently `process.cwd()` is added in `pluginThis`, causing it to be missing in actual `loadPaths`.
This PR modifies the logic to add `process.cwd()` in `adjustOptions` so that it affect actual `loadPaths`.

- Fixes https://github.com/sass/dart-sass/issues/1920

https://github.com/sass/sass-spec/pull/1893

### DIFF
--- a/lib/src/legacy/index.ts
+++ b/lib/src/legacy/index.ts
@@ -83,6 +83,10 @@ function adjustOptions<sync extends 'sync' | 'async'>(
     throw new Error('Either options.data or options.file must be set.');
   }
 
+  // In legacy API, the current working directory is always attempted before
+  // any load path.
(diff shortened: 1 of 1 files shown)


nex3:
Can you enable the corresponding JS API tests for this?

ntkme:
@nex3 Regression test added. Test is passing here, but not passing in the spec repo as it looks like spec repo does not have support of cloning linked PR for embedded host.

nex3:
Yeah, tests in the spec repo that add tests for the JS API should add "[skip sass-embedded]" to their description to avoid running tests against the base version.

nex3 (approved):

PR #214, opened 2023-04-09
fix: resolve.require could not find the target path

callqh:
```js
let path  = resolve.require('sass-embedded-darwin-arm64'); // not found
```
Because the algorithm is looking for the `main` field in the `package.json` file.

### DIFF
--- a/npm/darwin-arm64/package.json
+++ b/npm/darwin-arm64/package.json
@@ -5,6 +5,7 @@
   "repository": "sass/embedded-host-node",
   "author": "Google Inc.",
   "license": "MIT",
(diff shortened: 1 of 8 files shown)


ntkme:
This is not an expected usage and will break how embedded host works today.

See https://github.com/sass/embedded-host-node/blob/main/lib/src/compiler-path.ts

callqh:
> This is not an expected usage and may break how embedded host works today.
> 
> See https://github.com/sass/embedded-host-node/blob/main/lib/src/compiler-path.ts

got,thk!

PR #228, opened 2023-06-09
Bump the protocol version

nex3:


Goodwine (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sass-embedded",
   "version": "1.63.3",
-  "protocol-version": "2.0.0",
+  "protocol-version": "2.1.0",
   "compiler-version": "1.63.3",
```
Goodwine, line 5: Looks like this needs to be 1.63.4-dev? 🤔

nex3: I wonder if we should actually get rid of this test. Since we're automatically bumping this package's version as part of the release process, making sure it corresponds during development of new versions is likely to just be busywork.

PR #232, opened 2023-06-21
fix Invalid "exports" target "dist/lib/index.js"

patricklx:
error message:
Invalid "exports" target "dist/lib/index.js" defined in the package config; targets must start with "./"


patricklx:
ah, this is already covered by https://github.com/sass/embedded-host-node/pull/231

PR #236, opened 2023-06-29
Workaround for shelljs in windows to fix CI

Goodwine:
`shell.cp()` creates invalid symlinks when copying directories in Windows.
See: https://github.com/shelljs/shelljs/issues/198


.github/workflows/ci.yml
nex3, line 1: Unrelated changes (might want to consider disabling YAML autoformatting in your editor)

nex3 (approved):

PR #237, opened 2023-07-07
Add support for SassCalculation

jerivas:
https://github.com/sass/sass/pull/3622
https://github.com/sass/dart-sass/pull/1988

### DIFF
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -72,7 +72,7 @@ export class FunctionRegistry<sync extends 'sync' | 'async'> {
             )
           ),
           result => {
-            result = simplify(result) as any;
+            result = simplify(result) as types.Value;
             if (!(result instanceof Value)) {
               const name =
                 request.identifier.case === 'name'
@@ -136,11 +136,11 @@ function simplify(value: unknown): unknown {
     const simplifiedArgs = value.arguments.map(
       simplify
     ) as List<CalculationValue>;
-    if (value.name == 'calc') {
+    if (value.name === 'calc') {
       return simplifiedArgs.get(0);
     }
-    if (value.name == 'clamp') {
-      if (simplifiedArgs.size != 3) {
+    if (value.name === 'clamp') {
+      if (simplifiedArgs.size !== 3) {
         throw new Error('clamp() requires exactly 3 arguments.');
       }
       return SassCalculation.clamp(
@@ -149,10 +149,10 @@ function simplify(value: unknown): unknown {
         simplifiedArgs.get(2)
       );
     }
-    if (value.name == 'min') {
+    if (value.name === 'min') {
       return SassCalculation.min(simplifiedArgs);
     }
-    if (value.name == 'max') {
+    if (value.name === 'max') {
       return SassCalculation.max(simplifiedArgs);
     }
     // @ts-expect-error: Constructor is private, but we need a new instance here
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -137,6 +137,9 @@ function simplify(value: unknown): unknown {
       simplify
     ) as List<CalculationValue>;
     if (value.name === 'calc') {
+      if (simplifiedArgs.size !== 1) {
+        throw new Error('calc() requires exactly 1 argument.');
+      }
       return simplifiedArgs.get(0);
     }
     if (value.name === 'clamp') {
@@ -155,8 +158,7 @@ function simplify(value: unknown): unknown {
     if (value.name === 'max') {
       return SassCalculation.max(simplifiedArgs);
     }
-    // @ts-expect-error: Constructor is private, but we need a new instance here
-    return new SassCalculation(value.name, simplifiedArgs);
+    throw new Error(`Unknown calculation function: ${value.name}`);
   }
   if (value instanceof CalculationOperation) {
     return simplify(
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -158,15 +158,13 @@ function simplify(value: unknown): unknown {
     if (value.name === 'max') {
       return SassCalculation.max(simplifiedArgs);
     }
-    throw new Error(`Unknown calculation function: ${value.name}`);
+    throw new Error(`Unknown calculation function: ${value.name}.`);
   }
   if (value instanceof CalculationOperation) {
-    return simplify(
-      new CalculationOperation(
-        value.operator,
-        simplify(value.left) as CalculationValue,
-        simplify(value.right) as CalculationValue
-      )
+    return new CalculationOperation(
+      value.operator,
+      simplify(value.left) as CalculationValue,
+      simplify(value.right) as CalculationValue
     );
   }
   return value;
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -11,12 +11,6 @@ import * as proto from './vendor/embedded_sass_pb';
 import {PromiseOr, catchOr, compilerError, thenOr} from './utils';
 import {Protofier} from './protofier';
 import {Value} from './value';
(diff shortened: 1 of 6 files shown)


lib/src/value/calculations.ts
```diff
@@ -0,0 +1,157 @@
...
+          '`value` and `max` are both undefined, but `min` is not a SassString or CalculationInterpolation.'
+        );
+      }
+      const values = minString.split(',').map(s => {
+        const parsed = parseFloat(s.trim());
+        return isNaN(parsed) ? undefined : new SassNumber(parsed);
+      });
+      const error = new Error(
+        `Expected \`min\` to be a comma-separated list of numbers, got \`${min}\``
+      );
```
nex3, line 69: In retrospect, I think the proposal as written is both unclear and overzealous here. Sass itself allows a `clamp()` with fewer than three arguments as long as any of them is an unquoted string or interpolation, _regardless_ of what it contains. I'll update the spec.

lib/src/value/calculations.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export class CalculationOperation implements ValueObject {
+  readonly operator: CalculationOperator;
+  readonly left: CalculationValue;
+  readonly right: CalculationValue;
+
+  constructor(
+    operator: CalculationOperator,
+    left: CalculationValue,
+    right: CalculationValue
```
nex3, line 120: I believe you can automatically declare and assign these with:

```suggestion
    readonly operator: CalculationOperator,
    readonly left: CalculationValue,
    readonly right: CalculationValue
```

nex3 (changes requested):

lib/src/function-registry.ts
```diff
@@ -119,3 +126,46 @@ export class FunctionRegistry<sync extends 'sync' | 'async'> {
...
+      );
+    }
+    if (value.name === 'min') {
+      return SassCalculation.min(simplifiedArgs);
+    }
+    if (value.name === 'max') {
+      return SassCalculation.max(simplifiedArgs);
+    }
+    // @ts-expect-error: Constructor is private, but we need a new instance here
+    return new SassCalculation(value.name, simplifiedArgs);
```
jerivas, line 159: @nex3 I started implementing `simplify()`, but I realized that returning `SassCalculation` instances here will not automatically simplify the results back into primitive types as is the case in `dart-sass`. I presume we don't want to re-implement all the [logic](https://github.com/sass/dart-sass/blob/262961cb576d46f5518097c2c79c3e0dfb0e384d/lib/src/value/calculation.dart#L135-L159) from min, max, and clamp here, right? How should I tackle this?

nex3: The simplification should be handled by the embedded compiler. That is, the embedded host should send an _unsimplified_ calculation object back to the compiler, and then the compiler should use the existing Dart simplification logic to simplify the values on that end.

ntkme:
We also need to implement change in `protofier.ts` for conversion between proto type and JS type.

lib/src/function-registry.ts
```diff
@@ -119,3 +126,46 @@ export class FunctionRegistry<sync extends 'sync' | 'async'> {
...
+    if (value.name === 'min') {
+      return SassCalculation.min(simplifiedArgs);
+    }
+    if (value.name === 'max') {
+      return SassCalculation.max(simplifiedArgs);
+    }
+    throw new Error(`Unknown calculation function: ${value.name}.`);
+  }
+  if (value instanceof CalculationOperation) {
+    return new CalculationOperation(
```
jerivas, line 164: @nex3 in the [Dart implementation](https://github.com/oddbird/dart-sass/blob/e70e5e39f029d99d77397030988d7e6ee384acfb/lib/src/node/compile.dart#L248) of `simplify`, we can call `SassCalculation.operate` to collapse `CalculationOperation`s into plain values, but that method is not available in the host. Since we want to leverage as much as the Dart logic as possible, I'm hesitant to implement [`_operateInternal`](https://github.com/oddbird/dart-sass/blob/e70e5e39f029d99d77397030988d7e6ee384acfb/lib/src/value/calculation.dart#L182) in TS. Do you see another viable alternative?

lib/src/protofier.ts
```diff
@@ -144,6 +156,69 @@ export class Protofier {
...
+          right: this.protofyCalculationValue(value.right),
+        }),
+      };
+    } else if (value instanceof CalculationInterpolation) {
+      result.value = {case: 'interpolation', value: value.value};
+    } else if (value instanceof SassString) {
+      result.value = {case: 'string', value: value.text};
+    } else if (value instanceof SassNumber) {
+      // @ts-ignore
+      result.value = this.protofy(value).value;
```
jerivas, line 197: I couldn't get TS to see that when `value` is a `SassNumber` I should just be able to protofy it and get a valid return value. The full error is:

```
Type '{ value: Value_String; case: "string"; } | { value: Value_Number; case: "number"; } | { value: Value_RgbColor; case: "rgbColor"; } | { value: Value_HslColor; case: "hslColor"; } | ... 8 more ... | { ...; }' is not assignable to type '{ value: Value_Number; case: "number"; } | { value: string; case: "string"; } | { value: string; case: "interpolation"; } | { value: Value_Calculation_CalculationOperation; case: "operation"; } | { ...; } | { ...; }'.
  Type '{ value: Value_String; case: "string"; }' is not assignable to type '{ value: Value_Number; case: "number"; } | { value: string; case: "string"; } | { value: string; case: "interpolation"; } | { value: Value_Calculation_CalculationOperation; case: "operation"; } | { ...; } | { ...; }'.
    Type '{ value: Value_String; case: "string"; }' is not assignable to type '{ value: string; case: "string"; }'.
      Types of property 'value' are incompatible.
        Type 'Value_String' is not assignable to type 'string'.ts(2322)
```

It seems like there's a mismatch between TS's `string` type and `Value_String`, but it shouldn't be relevant in this case because we have narrowed down the type to `SassNumber`. Any pointers on how to fix it?

lib/src/protofier.ts
```diff
@@ -278,4 +356,101 @@ export class Protofier {
...
+  }
+
+  /** Converts `value` to its Sass representation. */
+  private deprotofyCalculationValue(
+    value: proto.Value_Calculation_CalculationValue
+  ): CalculationValue {
+    switch (value.value.case) {
+      case 'number':
+        // @ts-ignore
+        return this.deprotofy(value) as SassNumber;
```
jerivas, line 417: Same as my previous comments, but for deprotofying

jerivas:
@nex3 @ntkme this is ready for another review

lib/src/function-registry.ts
nex3: The expectation is that the host returns an _entirely unsimplified_ calculation AST to the compiler. The simplification will be handled by the compiler on its own. You probably don't need a `simplify()` function at all.

lib/src/protofier.ts
nex3: I think the simplest solution here would just be to factor out a `protofyNumber()` function and call that from both here and `protofy()`.

nex3 (changes requested):

lib/src/function-registry.ts
jerivas: Ah, sorry, I was way overcomplicating it 😅 Should be fixed after 8d2d5a0193238d16c7c0bd4f7a4192ff8de4f1c9

lib/src/protofier.ts
```diff
@@ -341,6 +341,13 @@ export class Protofier {
     }
   }
 
+  private deprotofyNumber(number: proto.Value_Number): SassNumber {
```
nex3, line 344: Please add a short doc comment for this.

nex3 (approved):

PR #246, opened 2023-09-15
Pass the containing URL to importers under some circumstances

nex3:
Closes #3247
See sass/sass#3688
See sass/dart-sass#2083
See sass/sass-spec#1936

### DIFF
--- a/tool/get-language-repo.ts
+++ b/tool/get-language-repo.ts
@@ -32,7 +32,7 @@ export async function getLanguageRepo(
   // Workaround for https://github.com/shelljs/shelljs/issues/198
   // This file is a symlink which gets messed up by `shell.cp` (called from
   // `utils.link`) on Windows.
-  shell.rm('build/sass/spec/README.md');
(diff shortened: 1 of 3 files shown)


tool/get-language-repo.ts
```diff
@@ -32,7 +32,7 @@ export async function getLanguageRepo(
   // Workaround for https://github.com/shelljs/shelljs/issues/198
   // This file is a symlink which gets messed up by `shell.cp` (called from
   // `utils.link`) on Windows.
-  shell.rm('build/sass/spec/README.md');
+  if (process.platform === 'win32') shell.rm('build/sass/spec/README.md');
```
jathak, line 35: Is there a reason why this is no longer necessary on Linux/Mac?

jathak (approved):

tool/get-language-repo.ts
nex3: It was only ever present as a workaround for a Windows issue in the first place (see the comment above).

PR #249, opened 2023-09-26
Update protocol-version to 2.2.0

ntkme:
The 1.68.0 release failed due to mismatched protocol-version. This PR fixes that.

On the other hand, I wonder if we should automatically update the protocol-version in package.json as part of the dart-sass release process when GitHub Action commits to this repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sass-embedded",
   "version": "1.68.0",
-  "protocol-version": "2.1.0",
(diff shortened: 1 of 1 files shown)


nex3:
> On the other hand, I wonder if we should automatically update the protocol-version in package.json as part of the dart-sass release process when GitHub Action commits to this repo.

Yes, I think this is the best approach. Otherwise I will _definitely_ forget this every time.

Issue #252, opened 2023-09-28
Add support for first-class mixins


nex3:
See https://github.com/sass/sass/issues/626

PR #253, opened 2023-09-29
add first class mixins

connorskees:
Tracking issue: https://github.com/sass/sass/issues/626
Protobuf changes: https://github.com/sass/sass/pull/3674
Spec tests: sass/sass-spec#1933
dart-sass changes: sass/dart-sass#2073

Closes https://github.com/sass/embedded-host-node/issues/252

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,9 +74,9 @@ These dependencies are made available in different ways depending on context.
 ### Local Development
 
 When developing locally, you can download all of these dependencies by running
-`npm run init`. This provides the following options for `compiler` (for the
-embedded compiler), `protocol` (for the embedded protocol), and `api` (for the
-JS API):
+`npm install` and then `npm run init`. This provides the following options for
+`compiler` (for the embedded compiler), `protocol` (for the embedded protocol),
+and `api` (for the JS API):
 
 * `--<type>-path`: The local filesystem path of the package to use. This is
   useful when doing local development on both the host and its dependencies at
@@ -84,6 +84,9 @@ JS API):
 
 * `--<type>-ref`: A Git reference for the GitHub repository of the package to
   clone.
+
+If developing locally, you will need to specify both the compiler and language.
+For example: `npm run init -- --compiler-path=dart-sass --language-path=language`.
 
 By default:
 
--- a/lib/src/protofier.ts
+++ b/lib/src/protofier.ts
@@ -24,7 +24,7 @@ import {
   CalculationOperation,
   CalculationOperator,
 } from './value/calculations';
-import { SassMixin } from './value/mixin';
+import {SassMixin} from './value/mixin';
 
 /**
  * A class that converts [Value] objects into protobufs.
@@ -329,7 +329,6 @@ export class Protofier {
 
       case 'compilerMixin':
         return new SassMixin(value.value.value.id);
-  
 
       case 'calculation':
         return this.deprotofyCalculation(value.value.value);
--- a/lib/src/protofier.ts
+++ b/lib/src/protofier.ts
@@ -24,6 +24,7 @@ import {
   CalculationOperation,
   CalculationOperator,
 } from './value/calculations';
+import { SassMixin } from './value/mixin';
 
 /**
  * A class that converts [Value] objects into protobufs.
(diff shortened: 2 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -74,9 +74,9 @@ These dependencies are made available in different ways depending on context.
 ### Local Development
 
 When developing locally, you can download all of these dependencies by running
-`npm run init`. This provides the following options for `compiler` (for the
-embedded compiler), `protocol` (for the embedded protocol), and `api` (for the
-JS API):
+`npm install` and then `npm run init`. This provides the following options for
+`compiler` (for the embedded compiler), `protocol` (for the embedded protocol),
+and `api` (for the JS API):
```
connorskees, line 79: The content change here is to add "`npm install` and then ". The rest of the diff is reflowing to 80 characters.

connorskees:
cc @nex3

lib/src/value/mixin.ts
```diff
@@ -0,0 +1,43 @@
...
+export class SassMixin extends Value {
+  /**
+   * If this mixin is defined in the compiler, this is the unique ID that the
+   * compiler uses to determine which mixin it refers to.
+   *
+   * This is marked as public so that the protofier can access it, but it's not
+   * part of the package's public API and should not be accessed by user code.
+   * It may be renamed or removed without warning in the future.
+   */
+  readonly id: number | undefined;
```
ntkme, line 19: I guess this code is copied from `SassFunction`? In `SassFunction`, `id` can be `undefined` because it support both a host function (with signature+callback) and a compiler function (with id). For `SassMixin`, we don't support  host mixin. Therefore, `id` should never be `undefined`.

lib/src/protofier.ts
```diff
@@ -322,6 +327,10 @@ export class Protofier {
           'The compiler may not send Value.host_function.'
         );
 
+      case 'compilerMixin':
+        return new SassMixin(value.value.value.id);
+  
+
```
ntkme, line 333: Nit, keep only 1 newline.

lib/src/value/mixin.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  constructor(id: number) {
+    super();
+    this.id = id;
+  }
+
+  equals(other: Value): boolean {
+    return this.id === undefined
+      ? other === this
+      : other instanceof SassMixin && other.id === this.id;
```
ntkme, line 29: This should be just `return other instanceof SassMixin && other.id === this.id;`.

connorskees:
I believe CI was failing because I missed adding a link to the dart-sass changes. That should now be resolved.

nex3 (approved):

connorskees:
I'm a bit confused about where the CI failure is coming from. Is there a place I'm missing updating? All relevant code touching `Value` should have an associated `assertMixin` method, and I'm linking to all relevant PRs in the PR description.

ntkme:
Maybe we need to add a `mixin.d.ts` to https://github.com/sass/sass/tree/main/js-api-doc/value ?

connorskees:
I made the change in https://github.com/sass/sass/pull/3674, but CI fails there I believe because the spec file is in the `accepted/` folder and not `spec/`.

ntkme:
The spec failure is due to the mismatch between `js-api-doc/` and `spec/js-api/`. They need to be modified at the same time to mirror the change.

connorskees:
Is the file in the `spec/js-api` not auto-generated?

ntkme:
> Is the file in the `spec/js-api` not auto-generated?

As for as I know they are not auto-generated, therefore the CI has a check to ensure the documentation matches the spec.

connorskees:
I got CI passing on the sass/sass PR if someone is able to re-run the CI for this PR

connorskees:
The JS API steps seem to pass now, but the static analysis step is failing with an error that looks unrelated to my changes. Is the failure spurious?

ntkme:
> The JS API steps seem to pass now, but the static analysis step is failing with an error that looks unrelated to my changes. Is the failure spurious?

Because this repository does not have `package-lock.json` checked in, the issue could be that one of the lint related dependency got a new version and it is breaking the test. For now I guess we will have to debug and find out what went wrong. 

For a more stable and producible CI runs I think we could check in `package-lock.json` and use `npm ci` instead of `npm install` in tests. The only downside is that dependabot will be more noisy on each dependency update, but on the upside this actually give us better test coverage for each dependency version change than what we currently have.

nex3:
https://github.com/sass/embedded-host-node/pull/255 should fix the static analysis issues.

PR #259, opened 2023-10-19
[Color 4] Add support for CSS Color Level 4.

jgerigmeyer:
Spec - https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces-js.d.ts.md

- https://github.com/sass/sass/pull/3718
- https://github.com/sass/dart-sass/pull/2117
- https://github.com/sass/sass-spec/pull/1949

### DIFF
--- a/lib/src/value/color.ts
+++ b/lib/src/value/color.ts
@@ -76,21 +76,25 @@ export type KnownColorSpace =
   | ColorSpaceRgb
   | ColorSpaceXyz;
 
+/** Polar color space names (HSL, HWB, LCH, and Oklch spaces). */
+type PolarColorSpace = ColorSpaceHsl | ColorSpaceHwb | ColorSpaceLch;
+
 /**
  * Methods by which two hues are adjusted when interpolating between polar
  * colors.
  */
-export type HueInterpolationMethod =
+type HueInterpolationMethod =
   | 'decreasing'
   | 'increasing'
   | 'longer'
   | 'shorter';
 
-type Channels = {
+type ChannelOptions = {
   [key in ChannelName]?: number | null;
 };
 
-function getColorSpace(options: Channels) {
+/** Legacy determination of color space by channel name. */
+function getColorSpace(options: ChannelOptions): KnownColorSpace {
   if (typeof options.red === 'number') {
     return 'rgb';
   }
@@ -103,42 +107,36 @@ function getColorSpace(options: Channels) {
   throw valueError('No color space found');
 }
 
-function emitColor4ApiDeprecation(name: string) {
-  console.warn(
-    `Deprecation [color-4-api]: \`${name}\` is deprecated; use \`channel\` instead.`
-  );
-}
-
-function emitNullAlphaDeprecation() {
-  console.warn(
-    'Deprecation [null-alpha]: ' +
-      'Passing `alpha: null` without setting `space` is deprecated.' +
-      '\n' +
-      'More info: https://sass-lang.com/d/null-alpha'
-  );
-}
-
-function NaNtoNull(val: number) {
+function NaNtoNull(val: number): number | null {
   return Number.isNaN(val) ? null : val;
 }
 
-function NaNtoZero(val: number) {
+function NaNtoZero(val: number): number {
   return Number.isNaN(val) ? 0 : val;
 }
 
-function assertClamped(val: number, min: number, max: number, name: string) {
+function assertClamped(
+  val: number,
+  min: number,
+  max: number,
+  name: string
+): number {
   return Number.isNaN(val) ? val : fuzzyAssertInRange(val, min, max, name);
 }
 
-function coordToRgb(val: number) {
+function coordToRgb(val: number): number {
   return val * 255;
 }
 
-function normalizeHue(val: number) {
+function normalizeHue(val: number): number {
   return positiveMod(val, 360);
 }
 
-function encodeSpaceForColorJs(space?: KnownColorSpace) {
+/**
+ * Normalize discrepancies between Sass color spaces and ColorJS color space
+ * ids.
+ */
+function encodeSpaceForColorJs(space?: KnownColorSpace): string | undefined {
   switch (space) {
     case 'rgb':
       return 'srgb';
@@ -152,6 +150,10 @@ function encodeSpaceForColorJs(space?: KnownColorSpace) {
   return space;
 }
 
+/**
+ * Normalize discrepancies between Sass color spaces and ColorJS color space
+ * ids.
+ */
 function decodeSpaceFromColorJs(space: string, isRgb = false): KnownColorSpace {
   switch (space) {
     case 'srgb':
@@ -170,14 +172,22 @@ function decodeSpaceFromColorJs(space: string, isRgb = false): KnownColorSpace {
 
 // @TODO For some spaces (e.g. Lab and Oklab), ColorJS only accepts `l` and not
 // `lightness` as a channel name. Maybe a bug?
-function encodeChannelForColorJs(channel: ChannelName) {
+/**
+ * Normalize discrepancies between Sass channel names and ColorJS channel ids.
+ */
+function encodeChannelForColorJs(channel: ChannelName): string {
   if (channel === 'lightness') return 'l';
   return channel;
 }
 
-// Implement our own check of channel name validity for a given space, because
-// ColorJS allows e.g. `b` for either `blue` or `blackness` or `b` channels.
-function checkChannelValid(channel: ChannelName, space: KnownColorSpace) {
+/**
+ * Implement our own check of channel name validity for a given space, because
+ * ColorJS allows e.g. `b` for either `blue` or `blackness` or `b` channels.
+ */
+function validateChannelInSpace(
+  channel: ChannelName,
+  space: KnownColorSpace
+): void {
   let valid = false;
   switch (space) {
     case 'rgb':
@@ -216,30 +226,85 @@ function checkChannelValid(channel: ChannelName, space: KnownColorSpace) {
   }
 }
 
+/** Determine whether the given space is a polar color space. */
+function isPolarColorSpace(space: KnownColorSpace): space is PolarColorSpace {
+  switch (space) {
+    case 'hsl':
+    case 'hwb':
+    case 'lch':
+    case 'oklch':
+      return true;
+    default:
+      return false;
+  }
+}
+
+/**
+ * Normalize between ColorJS coordinates (which use `NaN`) and Sass Color
+ * coordinates (which use `null`).
+ */
+function getCoordsFromColor(
+  coords: [number, number, number],
+  isRgb = false
+): [number | null, number | null, number | null] {
+  let newCoords: [number | null, number | null, number | null] = coords;
+  if (isRgb) {
+    newCoords = (newCoords as [number, number, number]).map(coordToRgb) as [
+      number,
+      number,
+      number,
+    ];
+  }
+  return (newCoords as [number, number, number]).map(NaNtoNull) as [
+    number | null,
+    number | null,
+    number | null,
+  ];
+}
+
+function emitColor4ApiDeprecation(name: string) {
+  console.warn(
+    `Deprecation [color-4-api]: \`${name}\` is deprecated; use \`channel\` instead.`
+  );
+}
+
+function emitNullAlphaDeprecation() {
+  console.warn(
+    'Deprecation [null-alpha]: ' +
+      'Passing `alpha: null` without setting `space` is deprecated.' +
+      '\n' +
+      'More info: https://sass-lang.com/d/null-alpha'
+  );
+}
+
+// @TODO remove this
 Color.defaults.precision = 15;
 
 /** A SassScript color. */
 export class SassColor extends Value {
+  // ColorJS color object
   private color: ColorType;
+  // Boolean indicating whether this color is in RGB format
   private isRgb = false;
+  // Boolean indicating whether this color has a missing `alpha` channel
   private alphaMissing = false;
-  private channel0Id: string;
-  private channel1Id: string;
-  private channel2Id: string;
-  private clone() {
-    let coords = this.color.coords;
-    if (this.space === 'rgb') {
-      coords = coords.map(coordToRgb) as [number, number, number];
-    }
+  // Names for the channels of this color
+  private channel0Id: ChannelName;
+  private channel1Id: ChannelName;
+  private channel2Id: ChannelName;
+  // Private method for cloning this as a new SassColor
+  private clone(): SassColor {
+    const coords = getCoordsFromColor(this.color.coords, this.space === 'rgb');
     return new SassColor({
       space: this.space,
       [this.channel0Id]: coords[0],
       [this.channel1Id]: coords[1],
       [this.channel2Id]: coords[2],
+      alpha: NaNtoNull(this.color.alpha),
     });
   }
 
-  constructor(options: Channels & {space?: KnownColorSpace}) {
+  constructor(options: ChannelOptions & {space?: KnownColorSpace}) {
     super();
 
     if (options.alpha === null && !options.space) {
@@ -398,25 +463,29 @@ export class SassColor extends Value {
     }
   }
 
-  /** `this`'s alpha channel. */
+  /** This color's alpha channel, between `0` and `1`. */
   get alpha(): number {
     return NaNtoZero(this.color.alpha);
   }
 
-  /** `this`'s color space. */
+  /** The name of this color's color space. */
   get space(): KnownColorSpace {
     return decodeSpaceFromColorJs(this.color.spaceId, this.isRgb);
   }
 
-  /** Whether `this` is in a legacy color space. */
+  /**
+   * A boolean indicating whether this color is in a legacy color space (`rgb`,
+   * `hsl`, or `hwb`).
+   */
   get isLegacy(): boolean {
     return ['rgb', 'hsl', 'hwb'].includes(this.space);
   }
 
-  /** The values of this color's channels (excluding the alpha channel), or
-   * `null` for [missing] channels.
+  /**
+   * A list of this color's channel values (excluding alpha), with [missing
+   * channels] converted to `null`.
    *
-   * [missing]: https://www.w3.org/TR/css-color-4/#missing
+   * [missing channels]: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#missing_color_components
    */
   get channelsOrNull(): List<number | null> {
     let coords = this.color.coords;
@@ -426,7 +495,12 @@ export class SassColor extends Value {
     return List(coords.map(NaNtoNull));
   }
 
-  /** The values of this color's channels (excluding the alpha channel). */
+  /**
+   * A list of this color's channel values (excluding alpha), with [missing
+   * channels] converted to `0`.
+   *
+   * [missing channels]: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#missing_color_components
+   */
   get channels(): List<number> {
     let coords = this.color.coords;
     if (this.space === 'rgb') {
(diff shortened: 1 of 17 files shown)


lib/src/value/color.ts
```diff
@@ -3,202 +3,299 @@
...
-        scaledWhiteness,
-        scaledBlackness
-      );
-      this.blueInternal = hwbToRgb(
-        scaledHue - 1 / 3,
-        scaledWhiteness,
-        scaledBlackness
+    if (options.alpha === null && !options.space) {
+      console.warn(
+        'Passing `alpha: null` without setting `space` is deprecated.\n\nMore info: https://sass-lang.com/d/null-alpha'
```
jgerigmeyer, line 137: What's the correct way to handle deprecations?

lib/src/value/color.ts
```diff
@@ -3,202 +3,299 @@
...
       );
     }
 
-    this.alphaInternal =
-      color.alpha === undefined
+    const space = options.space ?? getColorSpace(options);
+    // TODO(jgerigmeyer) What to do about `null` alpha?
```
jgerigmeyer, line 142: Check if colorjs.io uses `NaN` for missing components.

lib/src/value/color.ts
```diff
@@ -3,202 +3,299 @@
...
+   * `null` for [missing] channels.
    *
-   * This is an internal property that's not an official part of Sass's JS API,
-   * and may be broken at any time.
+   * [missing]: https://www.w3.org/TR/css-color-4/#missing
    */
-  get hasCalculatedHsl(): boolean {
-    return !!this.hueInternal;
+  get channelsOrNull(): List<number | null> {
+    // TODO(jgerigmeyer) What to do about `null` channels?
```
jgerigmeyer, line 292: Maybe check for `NaN`?

lib/src/value/color.ts
```diff
@@ -3,202 +3,299 @@
...
+      case 'rgb':
+      case 'srgb':
+      case 'srgb-linear':
+      case 'display-p3':
+      case 'a98-rgb':
+      case 'prophoto-rgb':
+      case 'rec2020':
+        this.color = getColor({
+          spaceId: space,
+          // TODO(jgerigmeyer) What to do about `null` or `undefined` channels?
```
jgerigmeyer, line 159: Use `NaN`?

ntkme:
This faces  a challenge like Calculation simplification, which we decided to not implement on the host side. Currently what I see from this PR is that we would need to implement the same feature twice, once in dart for dart-sass, and then once in pure js for this project. I think we should consider adding Color.change feature to embedded protocol, that is to let the host send an existing color and arguments to change, and then the embedded compiler returns the color after change. This way we don’t need implement the same logic twice or worry about potential drifts in different implementation.

lib/src/value/color.ts
```diff
@@ -3,202 +3,375 @@
...
+export type KnownColorSpace =
+  | ColorSpaceHsl
+  | ColorSpaceHwb
+  | ColorSpaceLab
+  | ColorSpaceLch
+  | ColorSpaceRgb
+  | ColorSpaceXyz;
+
+type Channels = {
+  [key in ChannelName]?: number | null;
```
jgerigmeyer, line 112: Do we need to be concerned with string percentages here?

nex3:
@ntkme I'm generally wary about asking embedded hosts to make round trips with the compiler just to do value operations, because it opens up a lot of difficulties both in terms of the API design (especially in languages with a strict distinction between synchrony and asynchrony) and in terms of performance (I'd like to avoid hidden cliffs where an operation that looks like it could be efficient is actually very slow). I think maybe the best solution here would just be to say that we don't mandate embedded hosts support all the color operations that the JS API does.

lib/src/value/color.ts
```diff
@@ -3,202 +3,372 @@
...
+  OKLab,
+  P3,
+  ProPhoto,
+  REC_2020,
+  XYZ_D50,
+  XYZ_D65,
+  sRGB,
+  sRGB_Linear,
+} from 'colorjs.io/fn';
+import type {PlainColorObject} from 'colorjs.io/types/src/color';
```
nex3, line 32: This looks like a load from a private internal library, and the type isn't documented in [the official API docs](https://colorjs.io/api/). I'm worried if we use this we'll get broken by implementation detail changes in the upstream library.

lib/src/value/color.ts
```diff
@@ -3,202 +3,372 @@
...
-  private saturationInternal?: number;
-  private lightnessInternal?: number;
-  private readonly alphaInternal: number;
-
-  constructor(color: RgbColor);
-  constructor(color: HslColor);
-  constructor(color: HwbColor);
-  constructor(color: RgbColor | HslColor | HwbColor) {
+  private color: PlainColorObject;
+  private format: 'rgb' | null = null;
```
nex3, line 135: I don't understand this field—how does this differ from `color.space`? When is it null?

lib/src/value/color.ts
nex3: I think `console.warn()` is probably fine for the embedded host. That's fairly standard for JS libraries, right?

lib/src/value/color.ts
```diff
@@ -3,202 +3,372 @@
...
-    if (this.redInternal === undefined) {
-      this.hslToRgb();
+    emitColor4ApiDeprecation('red');
+    try {
+      return get(this.color, 'red');
+    } catch (error) {
+      return get(to(this.color, 'srgb'), 'red');
     }
-    return this.redInternal!;
   }
```
nex3, line 269: Currently, these will return the channel as-is for non-legacy RGB colors, instead of being converted to legacy RGB and returning that value. That's probably _not_ desirable—legacy code that's using these getters is generally going to be code that's assuming everything is convertible to legacy RGB anyway. Similarly, converting to sRGB first won't solve this completely, because sRGB's channels are in [0, 1] but RGB's are in [0, 255]

The spec is vague on this point, but it should probably update the definitions for the legacy channel getters to explicitly mention that they convert colors to legacy spaces first.

lib/src/value/color.ts
```diff
@@ -3,202 +3,372 @@
...
    */
-  get hasCalculatedHsl(): boolean {
-    return !!this.hueInternal;
+  get channelsOrNull(): List<number | null> {
+    return List(this.color.coords.map(c => (Number.isNaN(c) ? null : c)));
+  }
+
+  /** The values of this color's channels (excluding the alpha channel). */
+  get channels(): List<number> {
+    return List(this.color.coords);
```
nex3, line 371: This needs to convert NaNs to 0. Probably most methods here do. Definitely something to keep an eye out for in the API tests.

nex3 (changes requested):

lib/src/value/color.ts
jgerigmeyer: Yeah, this was a first draft and at least needs renaming/documentation, if it doesn't change entirely. Basically ColorJS always treats `rgb` as a format of `srgb`, not its own color space. So any usage of `rgb` here in embedded-host will be handled internally as an `srgb` color, and will require manual mapping to scale 0-255 down to 0-1 (or vice versa).

jgerigmeyer: Adjusted in https://github.com/sass/embedded-host-node/pull/259/commits/a27e7b6b50d4082f47de0a4a3bb2a520026ddefc

lib/src/value/color.ts
jgerigmeyer: Adjusted in https://github.com/sass/embedded-host-node/pull/259/commits/a27e7b6b50d4082f47de0a4a3bb2a520026ddefc

lib/src/value/color.ts
jgerigmeyer: Spec adjusted in https://github.com/sass/sass/pull/3736, embedded host adjusted in https://github.com/sass/embedded-host-node/pull/259/files/3713dee14d58ffd14f6ce574b9a7c5bfc8504bbf..04010b6960e5c45fc3f5859655a79f91da3efc4c

lib/src/value/color.ts
jgerigmeyer: Adjusted in https://github.com/sass/embedded-host-node/pull/259/commits/04010b6960e5c45fc3f5859655a79f91da3efc4c

lib/src/value/color.ts
```diff
@@ -208,166 +348,113 @@ export class SassColor extends Value {
...
-      this.saturationInternal = (100 * delta) / (max + min);
-    } else {
-      this.saturationInternal = (100 * delta) / (2 - max - min);
     }
+    return (
+      fuzzyHashCode(coords[0]) ^
+      fuzzyHashCode(coords[1]) ^
+      fuzzyHashCode(coords[2]) ^
+      fuzzyHashCode(this.alpha)
+    );
```
jgerigmeyer, line 454: I see that the Dart Sass implementation also includes `space.hashCode` here -- I assume that's something we need here too? If so, how is it calculated?

jgerigmeyer:
@nex3 I think this could use another round of quick review -- thanks!

lib/src/value/color.ts
nex3: In Dart, `hashCode` is defined for every object and just defaults to being based on identity (like `operator==`). I believe ImmuableJS's [`hash()` function](https://immutable-js.com/docs/latest@main/hash()/) will hash strings based on contents.

jgerigmeyer: Addressed in https://github.com/oddbird/embedded-host-node/commit/133dd7b52e26489de72f55e7032c4cad211077e6

nex3 (commented):
Looks good at a quick glance. Remember to document all functions and include return types before the final.

lib/src/value/color.ts
```diff
@@ -3,371 +3,766 @@
...
   }
 
   /** `this`'s red channel. */
   get red(): number {
-    if (this.redInternal === undefined) {
-      this.hslToRgb();
+    emitColor4ApiDeprecation('red');
+    let val = NaNtoZero(coordToRgb(this.color.srgb.red));
+    if (this.isLegacy && this.space !== 'rgb') {
+      val = fuzzyRound(val);
```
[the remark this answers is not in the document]
jgerigmeyer: Ah, ok. Not sure why I thought they were only rounded when converting another legacy non-RGB space. Adjusted in https://github.com/oddbird/embedded-host-node/commit/71d5f203fc8f234cff9df446c972aa25776a3908

lib/src/value/color.ts
```diff
@@ -3,371 +3,926 @@
...
 import {
   fuzzyAssertInRange,
   fuzzyEquals,
+  fuzzyHashCode,
   fuzzyRound,
   positiveMod,
 } from './utils';
-import {hash} from 'immutable';
+import {List, hash} from 'immutable';
+import Color from 'colorjs.io/dist/color.legacy.cjs';
```
jgerigmeyer, line 15: @nex3 The default ColorJS build does not support node14 -- do we need to maintain support for that?

jgerigmeyer:
@nex3 I think this is ready for another round of review. AFAICT all of the API is implemented. I can't reproduce the failing tests locally -- something to do with the legacy color definition in the embedded host?

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+    default:
+      return false;
+  }
+}
+
+/**
+ * Normalize between ColorJS coordinates (which use `NaN`) and Sass Color
+ * coordinates (which use `null`).
+ */
+function getCoordsFromColor(
```
nex3, line 247: Can you clarify that this is specifically going from the colorjs representation to the Sass representation? The name and documentation are currently kind of ambiguous. (Passing in a colorjs color would also work.)

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+  coords: [number, number, number],
+  isRgb = false
```
nex3, line 249: Document this parameter.

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+
+function checkChangeDeprecations(
+  options: {
+    [key in ChannelName]?: number | null;
+  },
+  channels: ChannelName[]
+) {
+  if (options.alpha === null) {
+    emitNullAlphaDeprecation();
+  }
```
nex3, line 278: Style nit: Put this condition on one line. Same several places below.

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
-  private saturationInternal?: number;
-  private lightnessInternal?: number;
-  private readonly alphaInternal: number;
-
-  constructor(color: RgbColor);
-  constructor(color: HslColor);
-  constructor(color: HwbColor);
-  constructor(color: RgbColor | HslColor | HwbColor) {
+  // ColorJS color object
+  private color: ColorType;
```
nex3, line 319: Style nit: Empty lines between fields.

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+  // Boolean indicating whether this color is in RGB format
```
nex3, line 320: I think this warrants a bit more explanation about the difference between the RGB and sRGB color spaces.

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+  private isRgb = false;
+  // Boolean indicating whether this color has a missing `alpha` channel
```
nex3, line 322: Explain that this isn't supported by colorjs. Maybe also include a tracking issue.

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
     }
 
-    this.alphaInternal =
-      color.alpha === undefined
-        ? 1
-        : fuzzyAssertInRange(color.alpha, 0, 1, 'alpha');
+    // @TODO ColorJS doesn't seem to allow initial `alpha` to be missing?
+    if (this.alphaMissing) {
+      this.color.alpha = NaN;
+    }
```
nex3, line 489: Why are we tracking `this.alphaMissing` explicitly if we can just set `color.alpha` to `NaN`?

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+   * Returns a new color that's the result of converting this color to the
+   * specified `space`.
    */
-  get hasCalculatedHsl(): boolean {
-    return !!this.hueInternal;
+  toSpace(space: KnownColorSpace): SassColor {
+    if (space === this.space) return this;
+    const color = this.clone();
+    color._toSpaceInternal(space);
+    return color;
```
nex3, line 639: I prefer to avoid making value objects mutable if possible, even just internally—it opens up a huge can of worms when trying to track down subtle bugs. Can we add a secret constructor overload that just takes a color object instead?

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+   * Returns a boolean indicating whether this color is in-gamut (as opposed to
+   * having one or more of its channels out of bounds) for the specified
+   * `space`, or its current color space if `space` is not specified.
+   */
+  isInGamut(space?: KnownColorSpace): boolean {
+    return this.color.inGamut(encodeSpaceForColorJs(space));
+  }
+
+  // Internal helper which mutates the current color object.
+  _toGamutInternal(space?: KnownColorSpace): void {
```
nex3, line 652: This should be explicitly marked `private` instead of using the leading underscore.

lib/src/value/color.ts
nex3, line 319: Fields should be marked `readonly` where possible.

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+        `Expected \`weight\` between \`0\` and \`1\`; received \`${weight}\`.`
+      );
     }
+
+    // ColorJS inverses the `weight` argument, where `0` is `this` and `1` is
+    // `color2`.
+    const color = this.color.mix(color2.color, 1 - weight, {
+      space: encodeSpaceForColorJs(this.space),
+      hue: hueInterpolationMethod,
+    } as any); // @TODO Waiting on new ColorJS release to fix type defs
```
nex3, line 823: Do we have a tracking issue or PR for this?

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+        propertyIsSet(options.red) ||
+        propertyIsSet(options.green) ||
+        propertyIsSet(options.blue)
+      ) {
+        space = 'rgb';
+      }
+      if (space !== this.space) {
+        emitColor4ApiChangeSpaceDeprecation();
+      }
+    }
```
nex3, line 912: Can we factor this out into a helper?

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+
+    const color = this.toSpace(space);
+    const getChangedValue = (channel: ChannelName) => {
+      if (propertyIsSet(options[channel])) {
+        return options[channel];
+      }
+      return color.channel(channel);
+    };
+    let changedColor: SassColor;
```
nex3, line 936: Consider moving the switch below into a helper as well so you can just `return this.changeHelper(...).toSpace(...)`.

nex3 (changes requested):

lib/src/value/utils.ts
```diff
@@ -115,7 +115,7 @@ export function fuzzyAssertInRange(
   throw valueError(`${num} must be between ${min} and ${max}`, name);
 }
 
-/** Returns `dividend % modulus`, but always in the range `[0, modulus)`. */
```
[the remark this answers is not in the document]
jgerigmeyer: Good catch -- fixed in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653.

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+  if (typeof options.saturation === 'number') {
+    return 'hsl';
+  }
+  if (typeof options.whiteness === 'number') {
+    return 'hwb';
+  }
+  throw valueError('No color space found');
+}
+
+function NaNtoNull(val: number): number | null {
```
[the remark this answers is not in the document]
jgerigmeyer: Added documentation in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
```diff
@@ -3,371 +3,1112 @@
...
+    case 'p3':
+      return 'display-p3';
+    case 'prophoto':
+      return 'prophoto-rgb';
+  }
+  return space as KnownColorSpace;
+}
+
+// @TODO For some spaces (e.g. Lab and Oklab), ColorJS only accepts `l` and not
+// `lightness` as a channel name. Maybe a bug?
```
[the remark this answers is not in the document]
jgerigmeyer: Added issue/PR links in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Renamed and added comments in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Addressed in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Addressed in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Added comment and link in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Removed in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Reworked to use a constructor in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Addressed in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Addressed in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Done in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

lib/src/value/color.ts
jgerigmeyer: Done in https://github.com/sass/embedded-host-node/pull/259/commits/18ab10ee84ce3b3321b8a8b98f52c7fb79650653

jgerigmeyer:
@nex3 Thanks for the review -- I think I've addressed everything, and this is ready for another round!

nex3 (approved):

nex3:
Can you rebase this onto the `feature.color-4` branch?

jgerigmeyer:
> Can you rebase this onto the `feature.color-4` branch?

@nex3 Done, though I had already been basing this on `main` and it looks like `feature.color-4` is behind `main`, so there are some additional changes here until `main` is merged into `feature.color-4`.

PR #260, opened 2023-10-30
[Package Importer] Embedded Host

jamesnw:
https://github.com/sass/sass/issues/2739

Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/package-importer.d.ts.md) is accepted.

- [x]  dart-sass - https://github.com/sass/dart-sass/pull/2130
- [x]  sass - https://github.com/sass/sass/pull/3728
- [x]  sass-spec - https://github.com/sass/sass-spec/pull/1952


lib/src/importer-registry.ts
```diff
@@ -7,10 +7,19 @@ import {URL} from 'url';
...
+  FileImporter,
+  Importer,
+  NodePackageImporter,
+  Options,
+} from './vendor/sass';
 import * as proto from './vendor/embedded_sass_pb';
 import {catchOr, thenOr, PromiseOr} from './utils';
 
+export const nodePackageImporter: NodePackageImporter = {
+  _NodePackageImporterBrand: '',
```
stof, line 20: Maybe this should use a Symbol rather than an empty string (external code can create a property with an empty string as value but won't be able to create an object with the same symbol as value without reusing the exported object)

jamesnw: I'm open to making this value something else, but I think a Symbol would essentially add a limited deterrent, but not give us any additional assurance. The [check currently](https://github.com/sass/embedded-host-node/pull/260/files#diff-7f23a509904ce33b4c02b582e976900ecfd6f440b3ccbd08a7e8e7eb87e9ecbeR67-R68) is already able to differentiate the exact exported object from a user-created object that conforms to the NodePackageImporter type.

stof: The runtime check does. But the type checker would not if you create an object with such property.
when using a symbol, you won't be able to create an object with the same symbol (unless the symbol gets exported to allow that)

stof: I think a solution for that is to use the Symbol as key as well:

```ts
// sass exports
declare const NPISymbol: unique symbol;
export interface NodePackageImporterSymbolKey {
  [NPISymbol]: unknown;
}

export const nodePackageImporterSymbolKey: NodePackageImporterSymbolKey = {
  [NPISymbol]: NPISymbol
};
```

there, there is no way to access the symbol from the exported `sass.nodePackageImporter` (except by using `Object.getOwnPropertySymbols` and then guessing which symbol is the right one in that list, but the type checker will still not allow you to refer to that symbol to defined the type for this `fakeSymbolImporterTyped`).

jamesnw: I've updated the [Codesandbox](https://codesandbox.io/s/sass-package-importer-type-exploration-2r4sjy?file=/src/index.ts), and I think that still can be minimally evaded with a cast-

```ts
// sass exports
export const compileSymbolKey = (pkgImporter: NodePackageImporterSymbolKey) =>
  null;

// sass usage
const fakeSymbolKey = {} as sass.NodePackageImporterSymbolKey;
sass.compileSymbolKey(fakeSymbolKey);
```

My thought is that each of these are adding higher barriers to jump, more of a sign to a developer that they should not be doing this. If more deterrent is needed, an alternate tact would be to make it more explicit to the developer that they are doing something wrong, something like-

```ts
export const NodePackageImporterWarning: NodePackageImporterWarning = {
  _NodePackageImporterBrand:
    "Do not make your own instance. You must use the exported object."
};
```

This would be easier to fake, but harder for developers to expect it to work.

I'll defer to @nex3 on how much of a deterrent we want to make this.

stof: @jamesnw any time you use the `as` operator in typescript, you basically disable the type checker on that line (well, only partially as typescript still expects some overlap between the inferred type and the cast to reject full mistakes, requiring an intermediate cast to `unknown` if this is intended, but I could do `const fakeSymbolKey = true as unknown as sass.NodePackageImporterSymbolKey;` and be accepted).
You cannot both rely on the type checker to report invalid usages and telling it to assume that your code is valid...

nex3: I'm not _too_ worried about people working around the type checking here, but I think the `declare const symbol` approach more accurately communicates the intent anyway so I'm in favor of it.

jamesnw: I've made this change, with the spec change [here](https://github.com/sass/sass/pull/3748).

stof: ah, using the symbol itself as `sass.nodePackageImporter` is indeed even better

lib/src/importer-registry.ts
```diff
@@ -41,12 +48,35 @@ export class ImporterRegistry<sync extends 'sync' | 'async'> {
       );
   }
 
+  // Type predicate for NodePackageImporter
+  isNodePackageImporter(
+    importer: Importer<sync> | FileImporter<sync> | NodePackageImporter
+  ): importer is NodePackageImporter {
+    return typeof importer === 'symbol';
```
stof, line 55: this is wrong. You should not accept _any_ symbol as being the NodePackageImporter.

The check should be `importer === nodePackageImporter`

jamesnw: This is not the equality check- that happens within `register()`. This is (loosely) a type predicate, which allows us to differentiate between NodePackageImporter-like values and Importer/FileImporter-like values. 

But you're right this is unclear, and since we've simplified the type now, I can just move this check inside `register()`, which I did in https://github.com/sass/embedded-host-node/pull/260/commits/e591fadfa1911d887a81648e3b9b677703004c59. Thanks!

lib/src/legacy/index.ts
```diff
@@ -167,8 +171,24 @@ function convertOptions<sync extends 'sync' | 'async'>(
...
+    return {
+      ...convertedOptions,
+      importers: [nodePackageImporter, ...importers],
+    };
+  } else {
+    return {
+      ...convertedOptions,
+      importers,
+    };
+  }
```
nex3, line 191: Would it be possible to inject the importer into `importers` before returning the options, so you don't have to create a totally new copy of it here?

lib/src/importer-registry.ts
```diff
@@ -43,10 +53,31 @@ export class ImporterRegistry<sync extends 'sync' | 'async'> {
...
   register(
-    importer: Importer<sync> | FileImporter<sync>
+    importer: Importer<sync> | FileImporter<sync> | NodePackageImporter,
+    _entryPointURL?: string | URL
   ): proto.InboundMessage_CompileRequest_Importer {
     const message = new proto.InboundMessage_CompileRequest_Importer();
-    if ('canonicalize' in importer) {
+    if (typeof importer === 'symbol') {
+      if (importer !== nodePackageImporter) {
+        throw 'Incorrect Node Package Importer used';
```
nex3, line 62: This doesn't provide the user with much information about what was incorrect about it. I think it might make more sense to say `Unknown importer ${importer}`, since we don't know that they intended it to be a Node importer at all. Maybe a future version of Sass adds another symbol-based importer and they pass it to an older version of the API.

lib/src/importer-registry.ts
```diff
@@ -43,10 +53,31 @@ export class ImporterRegistry<sync extends 'sync' | 'async'> {
...
+      }
+      const importerMessage = new proto.NodePackageImporter();
+      let entryPointURL = _entryPointURL ?? require.main?.filename;
+      entryPointURL = entryPointURL?.toString();
+      if (entryPointURL === legacyImporterProtocol)
+        entryPointURL = require.main?.filename;
```
nex3, line 68: Multiline `if` statements always need brackets.

lib/src/importer-registry.ts
```diff
@@ -43,10 +53,31 @@ export class ImporterRegistry<sync extends 'sync' | 'async'> {
 
   /** Converts an importer to a proto without adding it to `this.importers`. */
   register(
-    importer: Importer<sync> | FileImporter<sync>
+    importer: Importer<sync> | FileImporter<sync> | NodePackageImporter,
+    _entryPointURL?: string | URL
```
nex3, line 57: Why does this have a leading underscore?

nex3 (changes requested):

lib/src/importer-registry.ts
jamesnw: Changed in b3f5414d3c366bed91a220089b2f201cec0c178a

Side note, unless the user provides a description for their Symbol, this will error with `Unknown importer Symbol()` since we won't have access to the importer name.

lib/src/importer-registry.ts
jamesnw: Addressed in b3f5414d3c366bed91a220089b2f201cec0c178a

jamesnw:
Review comments were addressed, but I'll wait to request re-review until after the entry point logic is ironed out.

lib/src/legacy/index.ts
jgerigmeyer: @jamesnw Do you recall the specific problem caused by the symbol losing its unique flag? I refactored this in https://github.com/sass/embedded-host-node/pull/260/commits/97b3685dbdcbc75f29e0424bd8ff6fd2b4063ff3 and it didn't seem to cause issues that I can see.

lib/src/utils.test.ts
```diff
@@ -5,7 +5,7 @@ describe('utils', () => {
   describe('pathToUrlString', () => {
     it('encode relative path like `pathToFileURL`', () => {
       const baseURL = pathToFileURL('').toString();
-      for (let i = 0; i < 128; i++) {
```
jgerigmeyer, line 8: This (temporarily) works around a Node regression documented in https://github.com/nodejs/node/issues/51167 and https://github.com/sass/embedded-host-node/pull/264

lib/src/legacy/index.ts
jamesnw: I think I was trying to construct the importers value outside of the return statement, but didn't try an inline ternary. If you assign a unique symbol to an intermediate const, you lose the unique flag, but it looks like it can be avoided this way.

jgerigmeyer: Ah, that makes sense!

lib/src/importer-registry.ts
```diff
@@ -11,6 +11,13 @@ import {FileImporter, Importer, Options} from './vendor/sass';
 import * as proto from './vendor/embedded_sass_pb';
 import {catchOr, thenOr, PromiseOr} from './utils';
 
+export class NodePackageImporter {
+  entryPointPath?: string;
```
nex3, line 15: This shouldn't be a public field—since it can't be private, it might be best to make it a local symbol.

lib/src/importer-registry.ts
```diff
@@ -43,10 +50,26 @@ export class ImporterRegistry<sync extends 'sync' | 'async'> {
...
-    if ('canonicalize' in importer) {
+    if (importer instanceof NodePackageImporter) {
+      const importerMessage = new proto.NodePackageImporter();
+      const entryPointPath = importer.entryPointPath
+        ? p.resolve(process.cwd(), importer.entryPointPath)
+        : require.main?.filename;
+
+      if (!entryPointPath) {
+        throw new Error(
+          'Node Package Importer requires access to a filesystem.'
```
nex3, line 64: Don't title-case "Node package importer", and provide more information about _why_ this is failing. Explain that `require.main.filename` is unset.

nex3 (approved):

lib/src/importer-registry.ts
jgerigmeyer: Fixed in https://github.com/oddbird/embedded-host-node/commit/b8e96d64c1343d73d3ab053c00f5c8b3f94b77bf

lib/src/importer-registry.ts
jgerigmeyer: Added in https://github.com/oddbird/embedded-host-node/commit/b8e96d64c1343d73d3ab053c00f5c8b3f94b77bf

ntkme:
Please also update `protocol-version` in `package.json`.

https://github.com/sass/embedded-host-node/blob/728c8fef2a3813ddc7bc6beb98067baee2554fe4/package.json#L4

jamesnw:
> Please also update `protocol-version` in `package.json`.
> 
> https://github.com/sass/embedded-host-node/blob/728c8fef2a3813ddc7bc6beb98067baee2554fe4/package.json#L4


Completed in https://github.com/sass/embedded-host-node/pull/260/commits/988fabb2f42276ea68f10989e5a2b7d7d2223d8f

lib/src/importer-registry.ts
```diff
@@ -11,10 +11,13 @@ import {FileImporter, Importer, Options} from './vendor/sass';
...
+const entryPointPathKey = Symbol();
+
 export class NodePackageImporter {
-  entryPointPath?: string;
+  [entryPointPathKey]?: string;
+
   constructor(entryPointPath?: string) {
-    this.entryPointPath = entryPointPath;
+    this[entryPointPathKey] = entryPointPath;
   }
```
nex3, line 21: Per the latest proposal, this should now throw an error eagerly if `entryPointPath` isn't passed and `require.main.filename` doesn't exist.

nex3 (changes requested):

lib/src/importer-registry.ts
jgerigmeyer: Addressed in https://github.com/oddbird/embedded-host-node/commit/f05aa50b54c51c5811deb27ca7cdf17cf6057576

lib/src/importer-registry.ts
```diff
@@ -11,23 +11,23 @@ import {FileImporter, Importer, Options} from './vendor/sass';
...
-  [entryPointPathKey]?: string;
+  [entryPointDirectoryKey]?: string;
 
-  constructor(entryPointPath?: string) {
-    entryPointPath = entryPointPath
-      ? p.resolve(process.cwd(), entryPointPath)
+  constructor(entryPointDirectory?: string) {
+    entryPointDirectory = entryPointDirectory
+      ? p.resolve(process.cwd(), entryPointDirectory)
       : require.main?.filename;
```
jgerigmeyer, line 22: @jamesnw Does this need to be `p.dirname(require.main?.filename)`?

jamesnw: Good catch- fixed in https://github.com/sass/embedded-host-node/pull/260/commits/be27f0b1987c695d7f08bb4fc9d429bc8660fbfa

lib/src/importer-registry.ts
```diff
@@ -11,6 +11,28 @@ import {FileImporter, Importer, Options} from './vendor/sass';
...
+
+export class NodePackageImporter {
+  [entryPointDirectoryKey]?: string;
+
+  constructor(entryPointDirectory?: string) {
+    entryPointDirectory = entryPointDirectory
+      ? p.resolve(entryPointDirectory)
+      : require.main?.filename
+      ? p.dirname(require.main?.filename)
+      : undefined;
```
nex3, line 24: Add parens around nested ternary expressions.

lib/src/importer-registry.ts
```diff
@@ -11,6 +11,28 @@ import {FileImporter, Importer, Options} from './vendor/sass';
 import * as proto from './vendor/embedded_sass_pb';
 import {catchOr, thenOr, PromiseOr} from './utils';
 
+const entryPointDirectoryKey = Symbol();
+
+export class NodePackageImporter {
+  [entryPointDirectoryKey]?: string;
```
nex3, line 17: I believe this can be `readonly` and non-nullable, since it's always set in the constructor (if it doesn't throw an error).

nex3 (approved):

lib/src/importer-registry.ts
jgerigmeyer: Prettier strips out parens around nested ternary where they don't impact the end result. Newer versions of Prettier add an [`experimentalTernaries` option](https://prettier.io/docs/en/options#experimental-ternaries), but 1) even that only adjusts the line breaks in this case, and doesn't allow wrapping parens, and 2) our version of Prettier here is locked by [gts](https://github.com/google/gts). Do you prefer wrapping parens here even if it requires an `// eslint-disable` comment?

nex3 (approved):

lib/src/importer-registry.ts
```diff
@@ -52,7 +52,11 @@ export class ImporterRegistry<sync extends 'sync' | 'async'> {
 
   constructor(options?: Options<sync>) {
     this.importers = (options?.importers ?? [])
-      .map(importer => this.register(importer))
+      .map(importer =>
+        this.register(
+          importer as Importer<sync> | FileImporter<sync> | NodePackageImporter
```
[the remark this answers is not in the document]
jgerigmeyer: @nex3 See my [earlier comment](https://github.com/sass/embedded-host-node/pull/260#discussion_r1476240717). Because we made `[entryPointDirectoryKey]` non-nullable in this file but it doesn't exist in the `NodePackageImporter` defined in the sass/sass repo (which is what `importers` expects), the two types no longer match. This cast resolves that, but honestly I'm not sure this change is better than just making it nullable here.

lib/src/importer-registry.ts
nex3: No, if Prettier insists on this I suppose it's fine.

lib/src/importer-registry.ts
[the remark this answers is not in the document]
nex3: Ah, sorry, it's so hard to see when there are unresolved comments in the GitHub interface.

lib/src/importer-registry.ts
nex3: I'd definitely prefer to have this readonly and non-nullable, since that's a direct representation of the actual state of the class. The static errors raise an interesting issue, though—because TypeScript checks types structurally, and the existing `NodePackageImporter` type from the Sass package has no properties, it will effectively allow _any_ object to be passed as an importer. It should probably have its own secret symbol property to avoid this, which will in turn mean that we will need a cast here no matter what.

jgerigmeyer: @nex3 Ah, ok -- that makes sense. So adding a private readonly non-nullable Symbol key to the definition in the Sass repo, correct?

PR #264, opened 2023-12-15
Use lts/* and lts/-n for node version; work around Node breaking change.

jgerigmeyer:
~~This is mostly serving as a PR to test whether the [failing Node tests](https://github.com/sass/embedded-host-node/actions/runs/7088513717/job/19522337184) are also failing on `main` -- though I'm seeing them fail locally on `main`, so I suspect it's unrelated to the changes in #260.~~

This updates Node versions in CI, fixes a few code style typos, and works around a [Node regression](https://github.com/nodejs/node/issues/51167).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ defaults:
 
 env:
   PROTOC_VERSION: 3.x
-  DEFAULT_NODE_VERSION: 20.x # If changing this, also change jobs.tests.strategy.matrix.node_version
 
 on:
   push:
@@ -22,7 +21,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: ${{ env.DEFAULT_NODE_VERSION }}
+          node-version: 'lts/*'
           check-latest: true
 
       - name: Check out the language repo
@@ -43,7 +42,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu, macos, windows]
-        node-version: [20.x, 18.x, 16.x] # If changing this, also change env.DEFAULT_NODE_VERSION
+        node-version: ['lts/*', 'lts/-1', 'lts/-2']
       fail-fast: false
 
     steps:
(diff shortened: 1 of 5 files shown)


ntkme (changes requested):

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu, macos, windows]
-        node-version: [18.x, 16.x, 14.x] # If changing this, also change env.DEFAULT_NODE_VERSION
+        node-version: [20.x, 18.x, 16.x] # If changing this, also change env.DEFAULT_NODE_VERSION
```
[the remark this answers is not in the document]
ntkme: E.g. `'lts/*'` (latest lts, this needs quote as `*` is special character) and `lts/-1`, `lts/-2`, `lts/-3`

lib/src/utils.test.ts
```diff
@@ -5,7 +5,7 @@ describe('utils', () => {
   describe('pathToUrlString', () => {
     it('encode relative path like `pathToFileURL`', () => {
       const baseURL = pathToFileURL('').toString();
-      for (let i = 0; i < 128; i++) {
+      for (let i = 33; i < 128; i++) {
```
jgerigmeyer, line 8: @nex3 Node versions 18.19.0 and 20.6.0 introduced a breaking change where `pathToFileURL` now trims trailing whitespace from the provided path. This is a workaround (skipping test for whitespace characters) while we await feedback on the [related issue](https://github.com/nodejs/node/issues/51167) I filed with Node.

Another option would be to adjust our implementation of `pathToUrlString` to similarly trim leading and trailing whitespace.

nex3: Nice work tracking this down! I think this workaround is fine for now, but can you add a comment linking to the tracking issue?

nex3 (approved):

lib/src/utils.test.ts
jgerigmeyer: Comment added in https://github.com/oddbird/embedded-host-node/commit/f9a28ed3debb6655e1e42a94afef902678b5d68f

PR #265, opened 2023-12-16
Support musl-libc and android

### LINKED ISSUE #263: Support Windows arm64 via the win32-x64 variant
Currently if you try and install this on Windows 11 Arm64 no native variant is installed, and at runtime you will get

```
[webpack-cli] Error: Embedded Dart Sass couldn't find the embedded compiler executable. Please make sure the optional dependency sass-embedded-win32-arm64 is installed in node_modules.
    at C:\project\node_modules\sass-embedded\lib\src\compiler-path.ts:52:9
    at Object.<anonymous> (C:\project\node_modules\sass-embedded\lib\src\compiler-path.ts:58:3)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
    at Object.require.extensions.<computed> [as .js] (C:\project\node_modules\ts-node\src\index.ts:1608:43)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Function.Module._load (node:internal/modules/cjs/loader:938:12)
    at Module.require (node:internal/modules/cjs/loader:1115:19)
    at require (node:internal/modules/helpers:130:18)
    at Object.<anonymous> (C:\project\node_modules\sass-embedded\lib\src\async-compiler.ts:9:1)
```

I've successfully used Dart Sass (native x64) [via Ruby](https://github.com/ntkme/sass-embedded-host-ruby) on Windows via emulation (ruby host uses [this logic](https://github.com/ntkme/sass-embedded-host-ruby/blob/b2380b96c2e398605f440aaaf1138825a1acd9d7/ext/sass/Rakefile#L262-L275)) which is still much faster than old variants.

Short of a full arm64 win32 compile of Dart Sass, could this project be changed to fall back to install and expect `win32-x64` when it otherwise would desire `win32-arm64`? Would such a PR be considered?

ntkme:
This PR adds support for musl-libc and android (and windows-arm64).

Because npm `optionalDependenices` only knows about generic linux, we have to install two variants on all linux system, and pick the correct one to run at runtime. It's not a huge problem other than waste some network bandwidth for download and a small amount of disk space.

Closes #263

### DIFF
--- a/lib/src/compiler-path.ts
+++ b/lib/src/compiler-path.ts
@@ -6,6 +6,10 @@ import * as fs from 'fs';
 import * as p from 'path';
 import {isErrnoException} from './utils';
 
+/**
+ * Detect if the current running node binary is linked with musl libc by
+ * checking if the binary contains a string like "/.../ld-musl-$ARCH.so"
+ */
 const isLinuxMusl = function () {
   return fs.readFileSync(process.execPath).includes('/ld-musl-');
 };
--- a/lib/src/compiler-path.ts
+++ b/lib/src/compiler-path.ts
@@ -12,21 +12,18 @@ const isLinuxMusl = function () {
 
 /** The full command for the embedded compiler executable. */
 export const compilerCommand = (() => {
-  let platform = process.platform as string;
-  let arch = process.arch;
-
-  if (platform === 'linux' && isLinuxMusl()) {
-    platform = 'linux-musl';
-  }
+  const platform =
+    process.platform === 'linux' && isLinuxMusl()
+      ? 'linux-musl'
+      : (process.platform as string);
 
   // https://github.com/sass/embedded-host-node/issues/263
   // Use windows-x64 emulation on windows-arm64
   //
   // TODO: Make sure to remove "arm64" from "npm/win32-x64/package.json" when
   // this logic is removed once we have true windows-arm64 support.
-  if (platform === 'win32' && arch === 'arm64') {
-    arch = 'x64';
-  }
+  const arch =
+    platform === 'win32' && process.arch === 'arm64' ? 'x86' : process.arch;
 
   // find for development
   for (const path of ['vendor', '../../../lib/src/vendor']) {
--- a/lib/src/compiler-path.ts
+++ b/lib/src/compiler-path.ts
@@ -6,14 +6,34 @@ import * as fs from 'fs';
 import * as p from 'path';
 import {isErrnoException} from './utils';
 
+const isLinuxMusl = function () {
+  return fs.readFileSync(process.execPath).includes('/ld-musl-');
+};
+
 /** The full command for the embedded compiler executable. */
 export const compilerCommand = (() => {
+  let platform = process.platform as string;
+  let arch = process.arch;
+
+  if (platform === 'linux' && isLinuxMusl()) {
+    platform = 'linux-musl';
+  }
+
+  // https://github.com/sass/embedded-host-node/issues/263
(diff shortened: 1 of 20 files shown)


ntkme:
The biggest problem here is we must install both -gnu and -musl at the same time, as that is a limitation of optional dependencies. For the embedded host it is not an issue, as we can detect the runtime and pick one to use, however it is a problem for the cli, as now we have two packages trying to install a bin with the same name.

```
  "bin": {
    "sass": "./dart-sass/sass"
  },
```

One way to solve this is to use a wrapper script to run the same platform detection logic and then invoke the dart sass command. The problem here is that node does not have proper support for `exec` call that replaces the current process, so that the node js wrapper need to stay alive until the child process exit, which is a minor performance issue.

nex3:
Do we need Node to have an `exec`? Since these packages are only ever going to be run on Linux anyway, could we have them ship a shell script to choose between them instead?

Also, does it make sense to have separate packages for glibc and musl, since they're both going to be installed anyway?

ntkme:
> Do we need Node to have an exec? Since these packages are only ever going to be run on Linux anyway, could we have them ship a shell script to choose between them instead?

It is unfortunate that the issue is closed without progress: https://github.com/nodejs/node/issues/21664

> Also, does it make sense to have separate packages for glibc and musl, since they're both going to be installed anyway?

We don't really need two packages, however the logic is simpler in this way because all the optional packages have the exact same structure.

nex3:
> It is unfortunate that the issue is closed without progress: https://github.com/nodejs/node/issues/21664

I mean, what if we didn't make the frontend script Node at all, and just made it a shell script instead?

ntkme:
> I mean, what if we didn't make the frontend script Node at all, and just made it a shell script instead?

I that case I'm not sure how to support posix shell and windows shell at the same time, unless we to with combined gnu/musl package, and have the shell script doing the detection work.

ntkme:
Or I we can still just separate packages, but include a separate bash script for the linux (-gnu) package, and use that for command line. For -musl, there will be no command script. I think this might be the best compromise.

nex3:
> I that case I'm not sure how to support posix shell and windows shell at the same time, unless we to with combined gnu/musl package, and have the shell script doing the detection work, in which way we can still go with the existing approach that each optional package contains one shell script.

If it's enough easier to have separate packages for gnu/musl, you could also just say "the gnu package is the only one with a script and that script determines which actual executable to load".

ntkme:
> If it's enough easier to have separate packages for gnu/musl, you could also just say "the gnu package is the only one with a script and that script determines which actual executable to load".

I also had the same idea and we almost posted at exact same time :-)

lib/src/compiler-path.ts
```diff
@@ -6,14 +6,34 @@ import * as fs from 'fs';
 import * as p from 'path';
 import {isErrnoException} from './utils';
 
+const isLinuxMusl = function () {
```
nex3, line 9: Nit: document this (in particular _why_ it works consistently)

nex3 (changes requested):

lib/src/compiler-path.ts
ntkme: Added a comment

nex3 (approved):

nex3:
Thanks!

ntkme:
@nex3 Any chance we can cut a release together with the current set of bug fixes, so that we can try it out and fix issues in the release pipeline if any.

PR #268, opened 2024-01-06
Serialize CSS string with SassString.prototype.toString()

ntkme:
This PR changes `SassString.prototype.toString()` to return a serialized CSS string similar to `dart-sass`.

Technically this is undocumented behavior, because it's not part of js-api spec. However, I have seen many users of ruby-sass, sassc-ruby, and dart-sass abusing this behavior, as demonstrated by the following example. Therefore, supporting this undocumented behavior would make adoption of `sass-embedded` more smooth.

``` js
import * as sass from 'sass-embedded';

const functions = {
  'image_path($source)': function (args) {
    const resolved = `/assets/${args[0].assertString('source').text}`;
    return new sass.SassString(resolved);
  },
  'image_url($source)': function (args) {
    const path = functions['image_path($source)'](args);
    return new sass.SassString(`url(${path.toString()})`, {quotes: false});
  },
};

console.log(
  // This is safe because it creates an AST node of `url()` Expression with SassString as argument. 
  sass.compileString('a { background: url(image_path("b.jpg")); }', {functions})
    .css
);

console.log(
  // This is not safe with the current `toString()` implementation, but users are tempted to do it
  // for convenience, as we don't have first class Expression support that would allow the host to
  // return a `url()` function expression.
  sass.compileString('a { background: image_url("b.jpg"); }', {functions}).css
);

const content = new sass.SassString('random #{$unsafe} content');
console.log(
  // In some cases this behavior is abused to the next level where string is directly embedded into
  // the sass input.
  // Note: `#` must to be escaped if sass interpolation is not desired.
  sass.compileString(
    `a { content: ${content.toString().replace(/#/g, '\\#')}; }`
  ).css
);
```

### DIFF
--- a/lib/src/value/string.ts
+++ b/lib/src/value/string.ts
@@ -126,8 +126,9 @@ export class SassString extends Value {
   }
 
   toString(): string {
-    if (!this.hasQuotes)
+    if (!this.hasQuotes) {
       return this.text.replace(/\0/g, '\uFFFD').replace(/\n */g, ' ');
+    }
 
     // https://drafts.csswg.org/cssom/#serialize-a-string
     let buffer = '"';
--- a/lib/src/value/string.ts
+++ b/lib/src/value/string.ts
@@ -126,7 +126,37 @@ export class SassString extends Value {
   }
 
   toString(): string {
-    return this.hasQuotes ? `"${this.text}"` : this.text;
+    if (!this.hasQuotes)
+      return this.text.replace(/\0/g, '\uFFFD').replace(/\n */g, ' ');
+
+    // https://drafts.csswg.org/cssom/#serialize-a-string
+    let buffer = '"';
+    for (const character of this.text) {
+      if (character === '\0') {
+        // If the character is NULL (U+0000), then the REPLACEMENT CHARACTER
+        // (U+FFFD).
+        buffer += '\uFFFD';
+      } else if (character === '\x22' || character === '\x5C') {
+        // If the character is '"' (U+0022) or "\" (U+005C), then the escaped
+        // character.
+        buffer += '\x5C' + character;
+      } else if (character < '\x20' || character === '\x7F') {
+        // If the character is in the range [\1-\1f] (U+0001 to U+001F) or is
+        // U+007F, then the character escaped as code point.
+        //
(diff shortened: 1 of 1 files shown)


lib/src/value/string.ts
```diff
@@ -126,7 +126,37 @@ export class SassString extends Value {
   }
 
   toString(): string {
-    return this.hasQuotes ? `"${this.text}"` : this.text;
+    if (!this.hasQuotes)
+      return this.text.replace(/\0/g, '\uFFFD').replace(/\n */g, ' ');
```
ntkme, line 130: This is effectively how unquoted string serialized by dart-sass.

lib/src/value/string.ts
```diff
@@ -126,7 +126,37 @@ export class SassString extends Value {
...
+        // and U+0061 to U+0066) to represent the code point in base 16,
+        // followed by a single SPACE (U+0020).
+        buffer += '\x5C' + character.codePointAt(0)!.toString(16) + '\x20';
+      } else {
+        // Otherwise, the character itself.
+        buffer += character;
+      }
+    }
+    buffer += '"';
+    return buffer;
```
ntkme, line 159: This is not exactly the same as how dart-sass would serialize quoted string, but it is a documented reference implementation. The end result is the same once the stylesheet is parsed by the browser.

The differences are:

1. dart-sass uses double quote or single quote depends on which kind of quote is present in the raw string. This implementation always use double quote as described by the linked csswg documentation.
2. dart-sass does not escape "TAB (U+0009)". This implementation escapes it as described by the linked csswg documentation.
3. dart-sass checks the next character and omits the trailing space if possible when escaping a character as code point. This implementation always produces s space as described by the linked csswg documentation.

jathak:
@ntkme: FYI, I've assigned Natalie to review this, though she's OOO this week

lib/src/value/string.ts
nex3, line 130: Style nit: use brackets for multi-line `if`s.

nex3 (commented):
I think this change is reasonable, but if we do make it consistent across all implementations we should also add it to the specification and the test suite. We should also consider adding similar requirements to other CSS-compatible value types.

PR #279, opened 2024-04-01
Implement the host side of the deprecations API

jathak:
See https://github.com/sass/sass/pull/3826
See https://github.com/sass/sass-spec/pull/1967
See https://github.com/sass/dart-sass/pull/2207


lib/src/deprecations.ts
ntkme, line 1: I think it would be better to not duplicate this information on every implementation. Either that we just don't expose the list of all deprecations as part of the JS API, or we implement a protocol to allow embedded host to send query to compiler and ask this information.

nex3 (changes requested):

lib/src/deprecations.ts
ntkme: I like the idea of maintain a text proto, which can generate ts definition and be directly used as argument in embedded-protocol.

jathak: Yeah, I think having a single representation and generating code for each implementation makes sense. I'd vote for writing it in YAML since it's a better authoring experience than JSON and I don't think the Dart protobuf library supports parsing text protos (https://github.com/google/protobuf.dart/issues/125).

Should this block releasing the deprecations API, or should we release it with just a simple synchronization test and then switch to generating the code afterwards?

nex3: I think the latter is fine.

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "sass-embedded",
-  "version": "1.72.0",
-  "protocol-version": "2.5.0",
-  "compiler-version": "1.72.0",
+  "version": "1.74.0",
```
nex3, line 3: You don't need to update the version and compiler-version here manually, it's done as part of the release process.

nex3 (approved):

package.json
jathak: Ah. I got the error for the protocol version and just updated all of them. Fixed.

PR #283, opened 2024-04-09
Add a deprecation for importer-without-url

### LINKED ISSUE #282: Add deprecation for base importer without a base URL
See https://github.com/sass/sass/issues/3831. All we need to do here is add this to the deprecations list, everything else will be handled by the embedded compiler.

nex3:
Closes #282
See sass/sass#3832
See sass/dart-sass#2213


nex3:
Per https://github.com/sass/dart-sass/issues/2208#issuecomment-2046013595, we've decided not to deprecate this after all.

PR #299, opened 2024-05-18
Parse Deprecation.forVersion on compiler side

ntkme:
https://github.com/sass/sass/pull/3868
https://github.com/sass/dart-sass/pull/2248


jathak (approved):
Remove the now-unnecessary `isActiveIn` function and `deprecations` import (see lints), and this LGTM

PR #300, opened 2024-05-21
Generate deprecations list from language repo

jathak:
See https://github.com/sass/sass/pull/3872.

### DIFF
--- a/tool/get-deprecations.ts
+++ b/tool/get-deprecations.ts
@@ -48,7 +48,7 @@ export async function getDeprecations(outDirectory: string) {
     tsText +=
       `  ${key}: {\n` +
       `    id: '${id}',\n` +
-      `    description: '${deprecation.description}'\n` +
+      `    description: '${deprecation.description}',\n` +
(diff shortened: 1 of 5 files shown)


tool/get-deprecations.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const yamlFile = 'build/sass/spec/deprecations.yaml';
+
+/**
+ * Converts a version string in the form X.Y.Z to be code calling the Version
+ * constructor, or null if the string is undefined.
+ */
+function toVersionCode(version: string | undefined): string | null {
+  if (!version) return null;
```
nex3, line 25: Since this is only ever used as a string value, I think it should return the string `'null'` here.

nex3 (changes requested):

tool/get-deprecations.ts
jathak: Done

nex3 (approved):

PR #309, opened 2024-07-27
Rewrite legacy API using containingUrl

### LINKED ISSUE #305: sass-embedded error with bootstrap 5.3.3
Since bootstrap 5.3.3, sass-embedded throw an error.
The problem was not present with bootstrap 5.3.2 and earlier, and is not present with normal sass package.

Here is the link to the open issue in bootstrap repo, with a reduced test case in a reproducible environment:
https://github.com/twbs/bootstrap/issues/39978#issue-2292514748

More specifically, the problem seems to be this line. If we remove it, everything runs fine:
https://github.com/twbs/bootstrap/blob/d2d4581790da2618d3fe063dafaa6205c73aabdd/scss/_variables.scss#L1751

ntkme:
This PR ports the strategy used by https://github.com/sass-contrib/sassc-embedded-shim-ruby to emulate SassC Ruby API and rewrites how legacy JS API is emulated.

The key is to use `containingUrl` to track the loading stack, and I'm able to get rid of significant amount of hacks in the previous implementation. In addition, it uses a pair of Importer and FileImporter, so that when LegacyImporter returns a file path, it's handled via FileImporter as a performance optimization to avoid reading large files on host side.

By the way, although I couldn't figure out what's wrong with #305 in the previous implementation, I did confirm locally that this reimplementation fixes that issue.

https://github.com/sass/sass/pull/3905
https://github.com/sass/dart-sass/pull/2291

- Closes #305.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,5 +5,19 @@
     // It would be nice to sort import declaration order as well, but that's not
     // autofixable and it's not worth the effort of handling manually.
     "sort-imports": ["error", {"ignoreDeclarationSort": true}],
+    // Match TypeScript style of exempting names starting with `_`.
+    // See: https://typescript-eslint.io/rules/no-unused-vars/
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      {
+        "args": "all",
+        "argsIgnorePattern": "^_",
+        "caughtErrors": "all",
+        "caughtErrorsIgnorePattern": "^_",
+        "destructuredArrayIgnorePattern": "^_",
+        "varsIgnorePattern": "^_",
+        "ignoreRestSiblings": true
+      }
+    ]
   }
 }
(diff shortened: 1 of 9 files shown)


nex3:
I believe this would break the ability of legacy importers to fully override `file:` URLs to do their own thing. For example:

```js
let count = 1;
console.log(sass.renderSync({
  data: '@import "file:foo"; @import "file:foo"',
  importer: function(url, prev) {
    console.log(url);
    return {contents: `foo {count: ${count++}}`}
  }
}).css.toString());
```

I'm also a little wary of dramatically changing a (mostly) stable implementation this close to the release of Dart Sass 2.0.0, where we're going to remove it entirely.

ntkme:
> I believe this would break the ability of legacy importers to fully override file: URLs to do their own thing.

Looks like this is already broken in the existing legacy implementation without this PR: the `sass` package will print `count: 1` and `count: 2`, but `sass-embedded` will print `count: 1` and `count: 1`.

I pushed another tiny change, and now with this PR it would print `count: 1` and `count: 2` just like `sass`.

PR #312, opened 2024-07-29
Support deprecation options in the legacy JS API

jathak:
See https://github.com/sass/sass/pull/3907.
See https://github.com/sass/sass-spec/pull/2006.
See https://github.com/sass/dart-sass/pull/2293.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,6 +1,7 @@
 const config = {
   preset: 'ts-jest',
   testEnvironment: 'node',
+  modulePathIgnorePatterns: ['<rootDir>/dart-sass/'],
 };
 
 export default config;
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,7 +1,7 @@
 const config = {
+  roots: ['<rootDir>/lib/', '<rootDir>/test/'],
   preset: 'ts-jest',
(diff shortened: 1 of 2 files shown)


lib/src/legacy/index.ts
```diff
@@ -156,7 +156,6 @@ function convertOptions<sync extends 'sync' | 'async'>(
           ),
         ]
       : undefined;
-
```
nex3, line 159: Nit: keep this

nex3 (approved):

nex3 (approved):

jest.config.ts
```diff
@@ -1,6 +1,7 @@
 const config = {
   preset: 'ts-jest',
   testEnvironment: 'node',
+  modulePathIgnorePatterns: ['<rootDir>/dart-sass/'],
```
ntkme, line 4: I had a similar but different commit to fix this issue: https://github.com/sass/embedded-host-node/pull/327/commits/01d7e42d3ae79d2413c61c1a359ad654524ffccd

Not sure which way is preferred.

jathak: Yeah, your solution looks better. I'll just add it to this PR for simplicity

PR #313, opened 2024-07-30
Add a Sass executable wrapper that forwards to the correct exe

nex3:
This allows users to run `npx sass` when they've just installed
`sass-embedded` and directly run the Dart VM.

### DIFF
--- a/bin/sass
+++ b/bin/sass
@@ -1,0 +1,62 @@
+#!/bin/bash -e
+
+dir="$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)"
+if [ "$(basename "$dir")" == bin ]; then
+    # Global executables are installed in .../bin/, and global libraries
+    # (including their dependencies) in .../lib/node_modules.
+    node_modules="$(dirname "$dir")/lib/node_modules"
+elif [ "$(basename "$dir")" == .bin ]; then
+    # Local dependency executables are installed in ./node_modules/.bin, and
+    # and their libraries in ./node_modules.
+    node_modules="$(dirname "$dir")"
+else
+    # If all else fails, try to find node_modules in a parent directory. We
+    # might execute this, for example, if someone runs the script from source.
+    while [ "$dir" != "/" ]; do
+        dir="$(dirname "$dir")"
+        if [ -d "$dir/node_modules" ]; then
+            node_modules="$dir/node_modules"
+            break
+        fi
+    done
+    echo "Could not find node_modules above ${BASH_SOURCE[0]}!" >&2
+    exit 1
+fi
+
+case "$(uname -s)" in
+    Linux*)
+        if [ ! -z "$ANDROID_ROOT" -a ! -z "$ANDROID_DATA" ]; then
+            os=android
+        else
+            os=linux
+        fi;;
+    Darwin*)
+        os=darwin;;
+    CYGWIN*|MINGW*|MSYS_NT*)
+        os=win32;;
+    *)
+        echo "Unknown operating system $(uname -s)!" >&2
+        exit 1
+esac
+
+musl=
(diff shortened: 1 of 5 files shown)


nex3:
cc @ntkme since this seems in your wheelhouse

ntkme:
Or maybe we just take a small bite in performance and create the wrapper as node JS script to use `child_process .execSync()`. Given that user probably won't call this extremely frequently, the extra node startup cost and memory cost might be not be too bad, and we can import and reuse the platform detect logic in https://github.com/sass/embedded-host-node/blob/main/lib/src/compiler-path.ts that way.

nex3:
Apparently Yarn doesn't support non-JS executables _at all_, so I guess it's really swimming upstream to try to do it this way. The performance hit is non-negligible—it looks like about 450ms of overhead, which will definitely add up if anyone tries running this in a tight loop. But it does seem like the infrastructure is just not there for doing this the faster way.

nex3:
Filed https://github.com/yarnpkg/berry/issues/6422 to track this on the Yarn end.

ntkme:
> it looks like about 450ms of overhead on my laptop,

Is that measuring `npm exec sass` or just running the script wrapper via node as `node sass`?

Either way, that is really slow. For reference, the overhead of invoking [ruby wrapper](https://github.com/sass-contrib/sass-embedded-host-ruby/blob/main/exe/sass) directly without using `bundle exec` is about 40ms on my M1 MBP. Running the ruby wrapper script under `bundle` as `bundle exec sass` has an additional 80ms overhead from `bundle`.

I haven't got a chance to measure it myself, but my guess is that even `npm exec`'s overhead is pretty significant.

nex3:
This is installing it via `npm install -g` and then just running `sass`. That doesn't invoke `npm` at all, just the symlinked script (which now runs `/usr/bin/env/node`). Poking around, it looks like some of that time is coming from loading dependencies through `utils.ts` that we don't actually use, but _most_ of it is the `isLinuxMusl()` function.

ntkme:
The current implementation of isLinuxMusl is very brute force that it reads the whole node binary into a buffer and then search for a substring which is unfortunately slow. A faster and more reliable way is to implement an ELF parser (only need bare minimum capability of parsing elf headers and program headers), extract the interpreter string, and check if `-musl` is in the basename of the interpreter string.

For reference, this is a Ruby implementation to extract the ELF interpreter string: https://github.com/sass-contrib/sass-embedded-host-ruby/blob/main/lib/sass/elf.rb

nex3:
I've filed https://github.com/sass/embedded-host-node/issues/314 to track making this check more efficient.

ntkme:
Shall we remove the `bin` section in the optional dependencies' `package.json` and the logic below if they don't actually work? 
https://github.com/sass/embedded-host-node/blob/9b5f08b016253b5e0edb03b153e8fc61346c32fc/tool/prepare-optional-release.ts#L115-L122

jathak (approved):

bin/sass
```diff
@@ -0,0 +1,24 @@
...
+// TODO npm/cmd-shim#152 and yarnpkg/berry#6422 - If and when the package
+// managers support it, we should make this a proper shell script rather than a
+// JS wrapper.
+
+try {
+child_process.execFileSync(
+  compilerCommand[0], [...compilerCommand.slice(1), ...process.argv.slice(2)], {
+    stdio: 'inherit',
+    windowsHide: true
+});
```
ntkme, line 15: Indent once?

nex3: This made me realize there's not a great way to get this file covered by the linter/formatter as-is, so I converted it to TypeScript.

jgerigmeyer:
@nex3 Is there a timeline for making a new release of `sass-embedded` that includes this fix?

nex3:
@jgerigmeyer We're planning to cut a release as soon as @jathak's work on making deprecations available to the legacy API lands.

PR #315, opened 2024-07-31
Improve the performance of detecting linux-musl platform by parsing ELF

### LINKED ISSUE #314: More efficiently check for musl libc
Currently, the check we run to determine whether the current Node process is running under musl libc—and thus which Linux binary to use for the embedded compiler—is very expensive. This is particularly problematic in light of https://github.com/sass/embedded-host-node/pull/313, which runs this check for every CLI invocation of the Sass compiler.

We should find a way to drive this cost down in order to minimize the overhead of starting the embedded host or the CLI. One option would be to port [the Ruby host's ELF-parsing logic](https://github.com/sass-contrib/sass-embedded-host-ruby/blob/main/lib/sass/elf.rb), although that's fairly complex. There are a few ELF parser libraries on npm, but none of them seem to support synchronous mode _and_ operating on a file descriptor rather than the whole executable, both of which are hard requirements here. It's possible there are other solutions, like looking for `/lib/libc.musl-x86_64.so.*`, but we'd have to do some research to verify that that's reliable.

ntkme:
Closes #314

### DIFF
--- a/lib/src/compiler-path.ts
+++ b/lib/src/compiler-path.ts
@@ -4,20 +4,22 @@
 
 import * as fs from 'fs';
 import * as p from 'path';
+import {getELFInterpreter} from './elf';
 import {isErrnoException} from './utils';
 
 /**
- * Detect if the current running node binary is linked with musl libc by
- * checking if the binary contains a string like "/.../ld-musl-$ARCH.so"
+ * Detect if the given binary is linked with musl libc by checking if
+ * the interpreter basename starts with "ld-musl-"
  */
-const isLinuxMusl = function () {
-  return fs.readFileSync(process.execPath).includes('/ld-musl-');
+const isLinuxMusl = function (path: string): boolean {
+  const interpreter = getELFInterpreter(path);
+  return p.basename(interpreter).startsWith('ld-musl-');
 };
 
 /** The full command for the embedded compiler executable. */
 export const compilerCommand = (() => {
   const platform =
-    process.platform === 'linux' && isLinuxMusl()
+    process.platform === 'linux' && isLinuxMusl(process.execPath)
       ? 'linux-musl'
       : (process.platform as string);
 
--- a/lib/src/compiler-path.ts
+++ b/lib/src/compiler-path.ts
@@ -4,17 +4,24 @@
 
 import * as fs from 'fs';
 import * as p from 'path';
-import {getELFInterpreter} from './elf';
+import {getElfInterpreter} from './elf';
 import {isErrnoException} from './utils';
 
 /**
  * Detect if the given binary is linked with musl libc by checking if
  * the interpreter basename starts with "ld-musl-"
  */
-const isLinuxMusl = function (path: string): boolean {
-  const interpreter = getELFInterpreter(path);
-  return p.basename(interpreter).startsWith('ld-musl-');
-};
+function isLinuxMusl(path: string): boolean {
+  try {
+    const interpreter = getElfInterpreter(path);
+    return p.basename(interpreter).startsWith('ld-musl-');
+  } catch (error) {
+    console.warn(
+      `Warning: Failed to detect linux-musl, fallback to linux-gnu: ${error.message}`
+    );
+    return false;
+  }
+}
 
 /** The full command for the embedded compiler executable. */
 export const compilerCommand = (() => {
--- a/lib/src/elf.ts
+++ b/lib/src/elf.ts
@@ -1,0 +1,99 @@
+// Copyright 2024 Google LLC. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import * as fs from 'fs';
+
+/**
+ * Read chunk of data from a file descriptor into a new Buffer.
+ */
+const readFileDescriptor = function (
+  fd: number,
+  position: number,
+  length: number
+): Buffer {
+  const buffer = Buffer.alloc(length);
+  let offset = 0;
+  while (offset < length) {
+    const bytesRead = fs.readSync(fd, buffer, {
+      offset: offset,
+      position: position + offset,
+    });
+    if (bytesRead === 0) {
+      throw new Error(`failed to read fd ${fd}`);
+    }
+
+    offset += bytesRead;
+  }
+  return buffer;
+};
+
+/**
+ * Parse an ELF file and return its interpreter.
+ */
+export const getELFInterpreter = function (path: string): string {
+  const fd = fs.openSync(path, 'r');
+  try {
+    const eIdent = new DataView(readFileDescriptor(fd, 0, 64).buffer);
+
+    if (
+      eIdent.getUint8(0) !== 0x7f ||
+      eIdent.getUint8(1) !== 0x45 ||
+      eIdent.getUint8(2) !== 0x4c ||
+      eIdent.getUint8(3) !== 0x46
+    ) {
+      throw new Error(`${path} is not an elf file.`);
+    }
+
+    const eIdentClass = eIdent.getUint8(4);
+    if (eIdentClass !== 1 && eIdentClass !== 2) {
+      throw new Error(`${path} has an invalid elf class.`);
+    }
+    const class32 = eIdentClass === 1;
+
+    const eIdentData = eIdent.getUint8(5);
+    if (eIdentData !== 1 && eIdentData !== 2) {
+      throw new Error(`${path} has an invalid endianness.`);
+    }
+    const littleEndian = eIdentData === 1;
+
+    const e_phoff = class32
+      ? eIdent.getUint32(28, littleEndian)
+      : Number(eIdent.getBigUint64(32, littleEndian));
+    const e_phentsize = class32
+      ? eIdent.getUint16(42, littleEndian)
+      : eIdent.getUint16(54, littleEndian);
+    const e_phnum = class32
+      ? eIdent.getUint16(44, littleEndian)
+      : eIdent.getUint16(56, littleEndian);
+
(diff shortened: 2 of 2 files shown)


lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
+// Copyright 2024 Google LLC. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import * as fs from 'fs';
+
+/**
+ * Read chunk of data from a file descriptor into a new Buffer.
+ */
+const readFileDescriptor = function (
```
nex3, line 10: Nit: for consistency:

```suggestion
function readFileDescriptor(
```

 I'll add a lint to enforce this.

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
+// Copyright 2024 Google LLC. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import * as fs from 'fs';
+
+/**
+ * Read chunk of data from a file descriptor into a new Buffer.
+ */
```
nex3, line 9: Nit:

```suggestion
/** Read a chunk of data from a file descriptor into a new Buffer. */
```

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    offset += bytesRead;
+  }
+  return buffer;
+};
+
+/**
+ * Parse an ELF file and return its interpreter.
+ */
```
nex3, line 33: Nit: 
```suggestion
/** Parse an ELF file and return its interpreter. */
```

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+export const getELFInterpreter = function (path: string): string {
```
nex3, line 34: Nit:

```suggestion
export function getElfInterpreter(path: string): string {
```

Even in JS code, we try to follow the Dart style of capitalizing 3+-letter acronyms as normal words.

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+  try {
+    const eIdent = new DataView(readFileDescriptor(fd, 0, 64).buffer);
+
+    if (
+      eIdent.getUint8(0) !== 0x7f ||
+      eIdent.getUint8(1) !== 0x45 ||
+      eIdent.getUint8(2) !== 0x4c ||
+      eIdent.getUint8(3) !== 0x46
+    ) {
+      throw new Error(`${path} is not an elf file.`);
```
nex3, line 45:
```suggestion
      throw new Error(`${path} is not an ELF file.`);
```

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    const eIdentClass = eIdent.getUint8(4);
+    if (eIdentClass !== 1 && eIdentClass !== 2) {
+      throw new Error(`${path} has an invalid elf class.`);
```
nex3, line 50:
```suggestion
      throw new Error(`${path} has an invalid ELF class.`);
```

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const eIdentData = eIdent.getUint8(5);
+    if (eIdentData !== 1 && eIdentData !== 2) {
+      throw new Error(`${path} has an invalid endianness.`);
+    }
+    const littleEndian = eIdentData === 1;
+
+    const e_phoff = class32
+      ? eIdent.getUint32(28, littleEndian)
+      : Number(eIdent.getBigUint64(32, littleEndian));
```
nex3, line 62: Probably worth adding a comment explaining why we believe it's safe to convert a bigint to a normal number here.

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+    const class32 = eIdentClass === 1;
+
+    const eIdentData = eIdent.getUint8(5);
+    if (eIdentData !== 1 && eIdentData !== 2) {
+      throw new Error(`${path} has an invalid endianness.`);
+    }
+    const littleEndian = eIdentData === 1;
+
+    const e_phoff = class32
```
nex3, line 60: Nit: Don't use underscores in variable names. Calling this `ePhoff` would be better, but since the fact that it's being used as an offset is relevant to the code, something explicit like `eProgHeaderOffset` would be even better.

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+      ? eIdent.getUint32(28, littleEndian)
+      : Number(eIdent.getBigUint64(32, littleEndian));
+    const e_phentsize = class32
+      ? eIdent.getUint16(42, littleEndian)
+      : eIdent.getUint16(54, littleEndian);
+    const e_phnum = class32
+      ? eIdent.getUint16(44, littleEndian)
+      : eIdent.getUint16(56, littleEndian);
+
+    const proghdrs = new DataView(
```
nex3, line 70: Similarly, calling this something less abbreviated like `progHeaders` would be clearer.

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+  return buffer;
+};
+
+/**
+ * Parse an ELF file and return its interpreter.
+ */
+export const getELFInterpreter = function (path: string): string {
+  const fd = fs.openSync(path, 'r');
+  try {
+    const eIdent = new DataView(readFileDescriptor(fd, 0, 64).buffer);
```
nex3, line 37: What does this `e` indicate here?

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const proghdrs = new DataView(
+      readFileDescriptor(fd, e_phoff, e_phentsize * e_phnum).buffer
+    );
+    for (let i = 0; i < e_phnum; i++) {
+      const byteOffset = i * e_phentsize;
+      const p_type = proghdrs.getUint32(byteOffset, littleEndian);
+      if (p_type !== 3) {
+        continue;
+      }
```
nex3, line 78: Nit: make `if` statements single-line if possible.

```suggestion
      if (p_type !== 3) continue;
```

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+      ? eIdent.getUint16(44, littleEndian)
+      : eIdent.getUint16(56, littleEndian);
+
+    const proghdrs = new DataView(
+      readFileDescriptor(fd, e_phoff, e_phentsize * e_phnum).buffer
+    );
+    for (let i = 0; i < e_phnum; i++) {
+      const byteOffset = i * e_phentsize;
+      const p_type = proghdrs.getUint32(byteOffset, littleEndian);
+      if (p_type !== 3) {
```
nex3, line 76: Add a brief comment explaining what "type 3" means.

lib/src/elf.ts
```diff
@@ -0,0 +1,99 @@
...
+        continue;
+      }
+
+      const p_offset = class32
+        ? proghdrs.getUint32(byteOffset + 4, littleEndian)
+        : Number(proghdrs.getBigUint64(byteOffset + 8, littleEndian));
+      const p_filesz = class32
```
nex3, line 83: What does the "z" indicate here?

lib/src/compiler-path.ts
```diff
@@ -4,20 +4,22 @@
...
 /**
- * Detect if the current running node binary is linked with musl libc by
- * checking if the binary contains a string like "/.../ld-musl-$ARCH.so"
+ * Detect if the given binary is linked with musl libc by checking if
+ * the interpreter basename starts with "ld-musl-"
  */
-const isLinuxMusl = function () {
-  return fs.readFileSync(process.execPath).includes('/ld-musl-');
+const isLinuxMusl = function (path: string): boolean {
+  const interpreter = getELFInterpreter(path);
```
nex3, line 15: I wonder if we should wrap this in a `try`/`catch`, print any error it produces to `stderr`, and just fall back to returning false if it errors. If the user is running musl, running the executable will error anyway, but if they're not that'll allow them to continue working without being strictly blocked on debugging why their ELF format isn't what we expect.

nex3 (changes requested):

lib/src/elf.ts
ntkme: In original kernel code this is named as `e_ident` (I think e stands for ELF):
 https://github.com/torvalds/linux/blob/21b136cc63d2a9ddd60d4699552b69c214b32964/include/uapi/linux/elf.h#L210C17-L210C24

lib/src/elf.ts
ntkme: This is the original variable name in linux kernel code: https://github.com/torvalds/linux/blob/21b136cc63d2a9ddd60d4699552b69c214b32964/include/uapi/linux/elf.h#L254

ntkme:
I mostly reused the variable names from the linux kernel code: https://github.com/torvalds/linux/blob/HEAD/include/uapi/linux/elf.h. Idea here is for whoever going to work on this code they can refer to the kernel API document and find the reference without too much effort: 

@nex3 Now I have renamed a few of them, and changed the snake case to camel case, let me know if you have any preferences on these naming.

nex3:
I think it's better to have explicit names than names that exactly match the kernel, especially since the explicit names are just expanded versions of the names from the kernel anyway so figuring out that they correspond won't be too hard for a reader. I'd also like to avoid single-letter affixes with unclear meaning like `e`, `p`, and `z`.

nex3 (approved):
Thanks!

ntkme:
@nex3 Okay, I've renamed everything to something that direct represent what they mean.

PR #323, opened 2024-08-05
Support linux-riscv64

ntkme:
It turns out while the commonly used https://github.com/nodesource/distributions do not support riscv64, nodejs itself does have riscv64 support, and distributions like debian sid and alpine linux do have nodejs for riscv64.

This PR adds riscv64 optional dependencies.


nex3:
Thanks! Can you open a PR against `dart-sass` as well to add a changelog entry for this?

PR #327, opened 2024-08-28
Fix race condition in SyncMessagePort

### LINKED ISSUE #326: Unexpected exception thrown due to race condition in SyncMessagePort
https://github.com/sass/dart-sass/pull/2312 made embedded compiler shutdown upon `ProtocolError` significantly faster, which revealed a race condition in SyncMessagePort that caused flaky test failures: https://github.com/sass/dart-sass/actions/runs/10552748019/job/29232018679

https://github.com/sass/embedded-host-node/blob/b54575994a4c51a0e6cfa3fbed236923609245b6/lib/src/sync-process/sync-message-port.ts#L127-L159

During shutdown the worker would post a `{"type":"exit","code":76}` event. For most of the cases, the main thread is fast enough that it's already passed the check of `Atomics.compareExchange(this.buffer, 0, BufferState.MessageSent, BufferState.AwaitingMessage) === BufferState.Closed`, and already waiting at `Atomics.wait(this.buffer, 0, BufferState.AwaitingMessage)`. In such case, the `exit` event is consumed correctly.

However, when the main thread is running slow, the port may have been closed before the check of `Atomics.compareExchange(this.buffer, 0, BufferState.MessageSent, BufferState.AwaitingMessage) === BufferState.Closed`, and by the time it runs the check, the error `the SyncMessagePort's channel is closed` is thrown, with the last `exit` event message not able to be consumed.

ntkme:
Fixes #326.

The idea of this PR is that instead of having only three states:

- AwaitingMessage = 0
- MessageSent = 1
- Closed = 2

This PR uses four states:

- AwaitingMessage =  0b00
- MessageSent = 0b01
- Closed & AwaitingMessage = 0b10
- Closed & MessageSent = 0b11

This allows `receiveMessage()` to correctly consume the very last message.

### DIFF
--- a/lib/src/sync-process/sync-message-port.ts
+++ b/lib/src/sync-process/sync-message-port.ts
@@ -23,18 +23,19 @@ enum BufferState {
    * the buffer to this state so that it can use `Atomics.wait()` to be notified
    * when it switches to `MessageSent`.
    */
-  AwaitingMessage,
+  AwaitingMessage = 0b00,
   /**
    * The state indicating that a message has been sent. Whenever an endpoint
    * sends a message, it'll set the buffer to this state so that the other
    * endpoint's `Atomics.wait()` call terminates.
    */
-  MessageSent,
+  MessageSent = 0b01,
   /**
-   * The state indicating that the channel has been closed. This never
-   * transitions to any other states.
+   * The bitmask indicating that the channel has been closed. This is masked on
+   * top of AwaitingMessage and MessageSent state. It never transitions to any
+   * other states once closed.
    */
-  Closed,
+  Closed = 0b10,
 }
 
 /**
@@ -158,13 +159,16 @@ export class SyncMessagePort extends EventEmitter {
     message = receiveMessageOnPort(this.port);
     if (message) return message.message;
 
-    assert.equal(Atomics.load(this.buffer, 0), BufferState.Closed);
+    assert.equal(
+      Atomics.and(this.buffer, 0, BufferState.Closed) & BufferState.Closed,
+      BufferState.Closed
+    );
     throw new Error("The SyncMessagePort's channel is closed.");
   }
 
   /** See `MessagePort.close()`. */
   close(): void {
-    Atomics.store(this.buffer, 0, BufferState.Closed);
+    Atomics.or(this.buffer, 0, BufferState.Closed);
     this.port.close();
   }
 }
--- a/lib/src/sync-process/sync-message-port.ts
+++ b/lib/src/sync-process/sync-message-port.ts
@@ -159,7 +159,9 @@ export class SyncMessagePort extends EventEmitter {
     message = receiveMessageOnPort(this.port);
     if (message) return message.message;
 
+    // Update the state to 0b10 after the last message is consumed.
     const oldState = Atomics.and(this.buffer, 0, BufferState.Closed);
+    // Assert the old state was either 0b10 or 0b11.
     assert.equal(oldState & BufferState.Closed, BufferState.Closed);
     throw new Error("The SyncMessagePort's channel is closed.");
   }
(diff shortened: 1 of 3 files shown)


lib/src/sync-process/sync-message-port.ts
```diff
@@ -158,13 +158,16 @@ export class SyncMessagePort extends EventEmitter {
     message = receiveMessageOnPort(this.port);
     if (message) return message.message;
 
-    assert.equal(Atomics.load(this.buffer, 0), BufferState.Closed);
+    assert.equal(
+      Atomics.and(this.buffer, 0, BufferState.Closed) & BufferState.Closed,
+      BufferState.Closed
+    );
```
ntkme, line 164: This now asserts that the old value is either 0x10 or 0x11 and sets the value to 0x10, so that if `receiveMessage()` is called one more time after close we can short-circuit at the first if statement.

lib/src/sync-process/sync-message-port.ts
```diff
@@ -23,18 +23,18 @@ enum BufferState {
    * the buffer to this state so that it can use `Atomics.wait()` to be notified
    * when it switches to `MessageSent`.
    */
-  AwaitingMessage,
+  AwaitingMessage = 0,
```
nex3, line 26: Can you write these as binary literals, since the bit position is now relevant?

lib/src/sync-process/sync-message-port.ts
```diff
@@ -23,18 +23,18 @@ enum BufferState {
...
   /**
    * The state indicating that a message has been sent. Whenever an endpoint
    * sends a message, it'll set the buffer to this state so that the other
    * endpoint's `Atomics.wait()` call terminates.
    */
-  MessageSent,
+  MessageSent = 1,
   /**
    * The state indicating that the channel has been closed. This never
    * transitions to any other states.
```
nex3, line 35: Document that this is used as a bit mask along with the `AwaitingMessage` and `MessageSent` states.

lib/src/sync-process/sync-message-port.ts
nex3: By "the first if statement", do you mean `if (Atomics.compareExchange(...))`? Since that compares both bits to `BufferState.Closed`, won't it only throw the error if the bits are precisely 0b10?

nex3 (changes requested):
Would it be possible to add a more direct test case for this to `sync-message-port.test.ts`? Or is the existing flaky failure the best we can do to exercise it?

lib/src/sync-process/sync-message-port.ts
ntkme: > By "the first if statement", do you mean `if (Atomics.compareExchange(...))`?

Yes.

> won't it only throw the error if the bits are precisely 0b10?

That is the desired behavior. Because if the state is 0b11, we want to receive the very last message that has already been sent right before close, and then set the state from 0b11 to 0b10 so that if `receiveMessage()` is called one more time, it throws as it's closed and we have no more messages.

ntkme:
> Would it be possible to add a more direct test case for this to sync-message-port.test.ts? Or is the existing flaky failure the best we can do to exercise it?

I've added a test as it turns out it is much more easier to produce it with a test case. On my local setup, the added test case 100% fails on main branch without adjusting timing artificially.

lib/src/sync-process/sync-message-port.ts
nex3: I see, the critical change here is that `Atomics.and()` is _updating_ the value. Putting a critical write operation in an `assert()` call is a bit confusing—although Node doesn't optimize out assertions by default, [users may configure it to](https://github.com/unassert-js/unassert), and readers may assume that the contents of an assertion aren't relevant to the correct behavior of the function.

I'd write this

```suggestion
    const oldState = Atomics.and(this.buffer, 0, BufferState.Closed);
    assert.equal(oldState & BufferState.Closed, BufferState.Closed);
```

...and probably also add a comment indicating why `Atomics.and()` is used here.

nex3 (approved):

Issue #170, opened 2022-12-13
Implement color spaces


nex3:
See https://github.com/sass/sass/issues/2831

jalberto:
This is breaking open-props https://github.com/argyleink/open-props/issues/546

nex3:
This issue should have been closed by https://github.com/sass/embedded-host-node/pull/330.

https://github.com/argyleink/open-props/issues/546 has nothing to do with the embedded host or with the color space JS API. As the error message indicates, the issue is that Sass doesn't think the `color()` function is allowed in plain CSS, when in fact it is.

jalberto:
you are right @nex3 thanks!

PR #330, opened 2024-09-12
Merge feature.color-4 into main

nex3:
See https://github.com/sass/embedded-host-node/issues/170
See https://github.com/sass/sass/pull/3930
See https://github.com/sass/sass-spec/pull/2013
See https://github.com/sass/dart-sass/pull/2341


lib/src/value/color.ts
```diff
@@ -3,371 +3,1182 @@
...
+  channels: ChannelName[]
+): void {
+  if (options.alpha === null) emitNullAlphaDeprecation();
+  for (const channel of channels) {
+    if (options[channel] === null) emitColor4ApiChangeNullDeprecation(channel);
+  }
+}
+
+/** Warn users about legacy color channel getters. */
+function emitColor4ApiGetterDeprecation(name: string): void {
```
jathak, line 306: All of the color-4-api deprecation warnings need some sort of handling for `silenceDeprecations` and `fatalDeprecations` since the host can't rely on the compiler to do the processing for any warnings emitted directly by the host.

jathak (changes requested):

lib/src/value/color.ts
nex3: I think it might be worth just merging this anyway so that we can get rid of the feature branch hassle, and then updating this along with your incoming change.

jathak (approved):

PR #331, opened 2024-09-13
Emit deprecation warnings for legacy JS API

jathak:
See https://github.com/sass/sass/pull/3932
See https://github.com/sass/sass-spec/pull/2016
See https://github.com/sass/dart-sass/pull/2343

### DIFF
--- a/lib/src/deprecations.ts
+++ b/lib/src/deprecations.ts
@@ -54,7 +54,6 @@ export function warnForHostSideDeprecation(
   deprecation: Deprecation,
   options?: DeprecationOptions
 ): void {
-  if (options) throw Error(message);
   if (
     deprecation.status === 'future' &&
     !isEnabledFuture(deprecation, options)
--- a/lib/src/deprecations.ts
+++ b/lib/src/deprecations.ts
@@ -2,7 +2,7 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-import {DeprecationOrId} from './vendor/sass';
+import {Deprecation, DeprecationOrId} from './vendor/sass';
 import {Version} from './version';
 
 export {deprecations} from './vendor/deprecations';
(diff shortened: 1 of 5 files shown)


lib/src/deprecations.ts
```diff
@@ -15,12 +15,83 @@ export {Deprecation, DeprecationOrId, DeprecationStatus} from './vendor/sass';
...
       return item;
     }
     return item.id;
   });
 }
+
+/**
+ * Shorthand for the subset of options related to deprecations.
+ */
+export type DeprecationOptions = {
```
nex3, line 31: Consider using [`Pick<>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) to derive this from the options type.

lib/src/deprecations.ts
```diff
@@ -15,12 +15,83 @@ export {Deprecation, DeprecationOrId, DeprecationStatus} from './vendor/sass';
...
+  fatalDeprecations?: (DeprecationOrId | Version)[];
+  futureDeprecations?: DeprecationOrId[];
+  silenceDeprecations?: DeprecationOrId[];
+};
+
+/**
+ * Handles a host-side deprecation warning, either emitting a warning, throwing
+ * and error, or doing nothing depending on the deprecation options used.
```
nex3, line 39:
```suggestion
 * an error, or doing nothing depending on the deprecation options used.
```

nex3 (changes requested):
Per our offline conversation, I think the best available solution for value API deprecations is to store a set of all active deprecation options and decide what to do based on those.

lib/src/deprecations.ts
```diff
@@ -25,53 +25,103 @@ export function getDeprecationIds(
   });
 }
 
+/**
+ * Map between active compilations and the deprecation options they use.
+ *
+ * This is used to determine which options to use when handling host-side
+ * deprecation warnings that aren't explicitly tied to a particular compilation.
+ */
+export const activeDeprecationOptions: Map<Object, DeprecationOptions> =
```
nex3, line 34: If you're using `Symbol`s as the key here, you could actually just use a `Record<Symbol, DeprecationOptions>`. Either way, the key type should be `Symbol`.

lib/src/deprecations.ts
```diff
@@ -25,53 +25,103 @@ export function getDeprecationIds(
...
+ *
+ * If no specific deprecation options are passed here, then options will be
+ * determined based on the options of the active compilations.
  */
 export function warnForHostSideDeprecation(
   message: string,
   deprecation: Deprecation,
   options?: DeprecationOptions
 ): void {
+  if (options) throw Error(message);
```
nex3: I don't understand this line. Is it left over from debugging?

lib/src/deprecations.ts
```diff
@@ -25,53 +25,103 @@ export function getDeprecationIds(
...
+  deprecation: Deprecation,
+  options?: DeprecationOptions
+): boolean {
+  if (!options) {
+    for (const potentialOptions of activeDeprecationOptions.values()) {
+      if (isSilent(deprecation, potentialOptions)) return true;
+    }
+    return false;
+  }
+  return getDeprecationIds(options?.silenceDeprecations ?? []).includes(
```
nex3, line 86:
```suggestion
  return getDeprecationIds(options.silenceDeprecations ?? []).includes(
```

There's no way for `options` to be undefined here. Also below.

nex3 (changes requested):

lib/src/deprecations.ts
jathak: Changed the key type to Symbol, but keeping this as a Map because it's better for adding/removing and checking the size

lib/src/deprecations.ts
jathak: Fixed

lib/src/deprecations.ts
jathak: Yes, though I'd removed it before requesting review, so I think this was just GitHub's review interface being bad

nex3 (approved):

PR #339, opened 2024-10-15
Fix error with deprecation warnings and legacy importers

jathak:
See https://github.com/sass/sass/pull/3898.
See https://github.com/sass/sass-spec/pull/2004.
See https://github.com/sass/dart-sass/pull/2282.

### DIFF
--- a/lib/src/legacy/utils.ts
+++ b/lib/src/legacy/utils.ts
@@ -8,7 +8,6 @@ import {pathToFileURL} from 'url';
 import {fileUrlToPathCrossPlatform} from '../utils';
 import {SourceSpan} from '../vendor/sass';
 import {legacyImporterFileProtocol} from './importer';
-import { remove } from 'immutable';
 
 /**
  * The URL protocol to use for URLs canonicalized using `LegacyImporterWrapper`.
--- a/lib/src/legacy/utils.ts
+++ b/lib/src/legacy/utils.ts
@@ -38,7 +38,13 @@ export function removeLegacyImporter(string: string): string {
 // syntax.
 export function removeLegacyImporterFromSpan(span: SourceSpan): SourceSpan {
   if (!span.url) return span;
-  return {...span, url: new URL(removeLegacyImporter(span.url.toString()))};
+  const url = removeLegacyImporter(span.url.toString());
+  return {
+    ...span,
+    url: URL.canParse(url)
+      ? new URL(url)
+      : new URL(url, `file://${process.cwd()}/`),
+  };
 }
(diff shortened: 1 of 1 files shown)


lib/src/legacy/utils.ts
```diff
@@ -38,7 +38,13 @@ export function removeLegacyImporter(string: string): string {
 // syntax.
 export function removeLegacyImporterFromSpan(span: SourceSpan): SourceSpan {
   if (!span.url) return span;
-  return {...span, url: new URL(removeLegacyImporter(span.url.toString()))};
+  const url = removeLegacyImporter(span.url.toString());
+  return {
+    ...span,
+    url: URL.canParse(url)
```
ntkme, line 44: `URL.canParse` requires `node >= 19.9.0`, so better do a `try` - `catch` instead.

ntkme (changes requested):

lib/src/legacy/utils.ts
```diff
@@ -38,7 +38,13 @@ export function removeLegacyImporter(string: string): string {
 // syntax.
 export function removeLegacyImporterFromSpan(span: SourceSpan): SourceSpan {
   if (!span.url) return span;
-  return {...span, url: new URL(removeLegacyImporter(span.url.toString()))};
+  const url = removeLegacyImporter(span.url.toString());
+  try {
+    return {...span, url: new URL(url)};
+  } catch (_) {
+    // Relative URL
+    return {...span, url: new URL(url, `file://${process.cwd()}`)};
```
nex3, line 46: Just shoving a pathname into a URL string isn't safe in general. Use Node's [`pathToFileUrl()`](https://nodejs.org/api/url.html#urlpathtofileurlpath-options).

lib/src/legacy/utils.ts
```diff
@@ -38,7 +38,13 @@ export function removeLegacyImporter(string: string): string {
 // syntax.
 export function removeLegacyImporterFromSpan(span: SourceSpan): SourceSpan {
   if (!span.url) return span;
-  return {...span, url: new URL(removeLegacyImporter(span.url.toString()))};
+  const url = removeLegacyImporter(span.url.toString());
+  try {
+    return {...span, url: new URL(url)};
+  } catch (_) {
```
nex3, line 44: If the URL is absolute, it won't use the basename anyway, so you can just pass it in eagerly rather than catching this error.

nex3 (changes requested):

nex3 (approved):

PR #342, opened 2024-10-22
Don't use `@import` in the legacy importer

### LINKED ISSUE #337: Export SassBoolean is missing.
The export for SassBoolean is undefined.

The `index.mjs` defines the export:
https://github.com/sass/embedded-host-node/blob/248437e15c67a352a9971c8cb3e9addbd8afc107/lib/index.mjs#L19

But it is not part of the export of `index.ts`:
https://github.com/sass/embedded-host-node/blob/248437e15c67a352a9971c8cb3e9addbd8afc107/lib/index.ts

This breaks usage of `sass-embedded` as drop-in replacement, if my code is using SassBoolean.

### LINKED ISSUE #340: Using legacy APIs with importers triggers `@import` deprecation warning
Even if the code doesn't contain `@import`, [the `@import` deprecation warning](https://sass-lang.com/documentation/breaking-changes/import/) appears.

```
Deprecation Warning: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import

  ╷
5 │ ;@import "sass-embedded-legacy-load-done:0";
  │          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
```

While I can set `silenceDeprecations`, it's a bit confusing.

Reproduction:
1. clone https://github.com/sapphi-red-repros/sass-embedded-legacy-importer-triggers-import-deprecated-warning
2. run `npm i`
3. run `node index.js`

nex3:
This unfortunately may make error reports look a little odd for errors
on the first line of SCSS files loaded via the legacy importer and
it'll give indented-syntax errors an off-by-one error, but there's an
easy fix: stop using the legacy API.

Closes #340

### DIFF
--- a/lib/src/legacy/importer.ts
+++ b/lib/src/legacy/importer.ts
@@ -46,7 +46,7 @@ export const endOfLoadProtocol = 'sass-embedded-legacy-load-done:';
 export const legacyImporterFileProtocol = 'legacy-importer-file:';
 
 /**
- * A random namespace for `sas:meta`, so we can use `meta.load-css()` at the end
+ * A random namespace for `sass:meta`, so we can use `meta.load-css()` at the end
  * of the file to signal that a load has finished without polluting a namespace
  * a user might actually use.
  */
--- a/lib/src/legacy/importer.ts
+++ b/lib/src/legacy/importer.ts
@@ -45,6 +45,13 @@ export const endOfLoadProtocol = 'sass-embedded-legacy-load-done:';
  */
 export const legacyImporterFileProtocol = 'legacy-importer-file:';
 
+/**
+ * A random namespace for `sas:meta`, so we can use `meta.load-css()` at the end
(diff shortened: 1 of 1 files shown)


lib/src/legacy/importer.ts
```diff
@@ -45,6 +45,13 @@ export const endOfLoadProtocol = 'sass-embedded-legacy-load-done:';
  */
 export const legacyImporterFileProtocol = 'legacy-importer-file:';
 
+/**
+ * A random namespace for `sas:meta`, so we can use `meta.load-css()` at the end
```
jathak, line 49:
```suggestion
 * A random namespace for `sass:meta`, so we can use `meta.load-css()` at the end
```

jathak (approved):

PR #344, opened 2024-10-26
Implement `sass --embedded` in pure JS mode

ntkme:
https://github.com/sass/dart-sass/pull/2413


ntkme:
I managed to use the isolate dispatcher from dart implementation so this is no longer needed.

bin/sass.ts
```diff
@@ -12,6 +13,10 @@ try {
     compilerCommand[0],
     [...compilerCommand.slice(1), ...process.argv.slice(2)],
     {
+      // Node blocks launching .bat and .cmd without a shell due to CVE-2024-27980
+      shell: ['.bat', '.cmd'].includes(
+        path.extname(compilerCommand[0]).toLowerCase()
+      ),
```
ntkme, line 19: This is to fix local test failure: https://github.com/sass/dart-sass/actions/runs/11563285433/job/32186274921

Basically the same thing as https://github.com/sass/embedded-host-node/pull/286

ntkme:
Because this PR assumes that the dartvm and dart2js sass-compiler will behave exactly same from an external viewpoint, this PR only run test the against dartvm in the CI. The dart2js based CI test is configured in the dart-sass repo itself.

Therefore, this PR has no dependency on dart-sass side of PR, and can be merged independently. - I would like to have this PR merged first, so that I don't have to keep resolving merge conflict every time the version number is bumped.

Of course, if we merge this PR first before merging the dart-sass PR, the API mode won't work. However, at least it allows users to run `npm exe sass-embedded` consistently on all platforms without native dart support.

@nex3 With the above said, it would be nice if you can take a look and merge this, even if we don't merge dart-sass PR right away.

tool/get-embedded-compiler.ts
```diff
@@ -14,7 +16,7 @@ import * as utils from './utils';
  * at `path`. By default, checks out the latest revision from GitHub.
  */
 export async function getEmbeddedCompiler(
-  outPath: string,
+  js?: boolean,
```
nex3, line 19: This should be in options, so that it's always passed by name. Same below.

tool/get-embedded-compiler.ts
```diff
@@ -14,7 +16,7 @@ import * as utils from './utils';
  * at `path`. By default, checks out the latest revision from GitHub.
  */
```
nex3, line 17: Please document the `js` option.

nex3 (approved):
I'm on board to land this first.

> However, at least it allows users to run npm exe sass-embedded consistently on all platforms without native dart support.

I don't completely understand this. It looks like this PR only affects local development. How will it affect what users can do at all?

ntkme:
Code review feedbacks have been applied.

ntkme:
> Wouldn't that require us to start shipping a "binary" package that adds a dependency on `sass`?

I added sass directly as an optional dependency. We can create an “binary” package that simply depends on “sass”, but due to how npm select optional dependencies, technically we need two separate packages to cover all systems:

- `{ "os": ["!android", "!darwin", "!linux", "!win32"] }`
- `{ "cpu": ["!arm", "!arm64", "!riscv64", "!x64"] }`

Let me know if that if preferred over having “sass” itself as direct optional dependency.

nex3:
Since `sass` is fairly heavyweight in terms of filesize, I think I'd rather have separate "binary" packages.

npm/all-unknown/README.md
nex3, line 1: Since these packages behave unusually relative to the binary packages, they should probably have slightly more descriptive READMEs and npm descriptions.

nex3 (approved):

npm/all-unknown/README.md
ntkme: Updated. Feel free to further amend the language if you have any other preference.

ntkme:
Note, here is the PR to update the dart-sass side of CI pipeline for this: https://github.com/sass/dart-sass/pull/2609

PR #349, opened 2024-11-04
Stricter checks in `LegacyColor` constructor

### LINKED ISSUE #348: `LegacyColor` has incorrect RGB values when either green or blue are zero
For example, the hex code `#ff0000` is just red, no green or blue, and is equivalent to `rgb(255, 0, 0)`. Constructing an instance of `LegacyColor` with these values results in `r == 0, g == 0, b == 255`:

```ts
new LegacyColor(255, 0, 0)
/*
sass.types.Color {
  inner: SassColor {
    redInternal: 0,
    greenInternal: 0,
    blueInternal: 255,
    alphaInternal: 1
  }
}
*/
```

I'm pretty sure [this line is the culprit](https://github.com/sass/embedded-host-node/blob/main/lib/src/legacy/value/color.ts#L25). If either `green` or `blue` are falsy, which `0` is, then the code acts as if the `constructor(argb: number)` variant was called. I think these should be stricter checks so that `0` doesn't trigger the wrong logical branch.

I understand that this is a deprecated API but I'd love to see it be fixed. I'd be happy to put together a pull request if you'd like.

Thanks in advance!

mrdziuban:
Fixes #348 

Rather than just checking if `green` or `blue` are falsy, the code now checks if they're `undefined` or `null`.

### DIFF
--- a/lib/src/legacy/value/color.ts
+++ b/lib/src/legacy/value/color.ts
@@ -2,12 +2,9 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+import {isNullOrUndefined} from '../../utils';
 import {SassColor} from '../../value/color';
 import {LegacyValueBase} from './base';
-
-function isUndefinedOrNull<A>(a: A): boolean {
-  return a === undefined || a === null;
-}
 
 export class LegacyColor extends LegacyValueBase<SassColor> {
   constructor(red: number, green: number, blue: number, alpha?: number);
@@ -26,7 +23,7 @@ export class LegacyColor extends LegacyValueBase<SassColor> {
     }
 
     let red: number;
-    if (isUndefinedOrNull(green) || isUndefinedOrNull(blue)) {
+    if (isNullOrUndefined(green) || isNullOrUndefined(blue)) {
       const argb = redOrArgb as number;
       alpha = (argb >> 24) / 0xff;
       red = (argb >> 16) % 0x100;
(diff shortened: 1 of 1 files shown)


lib/src/legacy/value/color.ts
```diff
@@ -22,7 +22,7 @@ export class LegacyColor extends LegacyValueBase<SassColor> {
     }
 
     let red: number;
-    if (!green || !blue) {
+    if (green == undefined || blue == undefined) {
```
ntkme, line 25: Use triple equal instead of double equal:

```suggestion
    if (green === undefined || blue === undefined) {
```

ntkme (changes requested):

lib/src/legacy/value/color.ts
mrdziuban: Got it, just pushed a change to explicitly check for both `undefined` and `null`

lib/src/legacy/value/color.ts
```diff
@@ -5,6 +5,10 @@
 import {SassColor} from '../../value/color';
 import {LegacyValueBase} from './base';
 
+function isUndefinedOrNull<A>(a: A): boolean {
+  return a === undefined || a === null;
+}
```
nex3, line 10: Please document this, and put it in `utils.ts`

lib/src/legacy/value/color.ts
```diff
@@ -5,6 +5,10 @@
 import {SassColor} from '../../value/color';
 import {LegacyValueBase} from './base';
 
+function isUndefinedOrNull<A>(a: A): boolean {
```
nex3, line 8:
```suggestion
function isUndefinedOrNull(a: unknown): boolean {
```

nex3 (changes requested):

nex3 (approved):

PR #352, opened 2024-12-10
Update relative pathToFileURL to match node

ntkme:
Node recently changed what is escaped and what is not in pathToFileURL, causing tests to fail on node 22 LTS and higher:
https://github.com/nodejs/node/pull/54545

### DIFF
--- a/lib/src/utils.ts
+++ b/lib/src/utils.ts
@@ -104,7 +104,7 @@ export function pathToUrlString(path: string): string {
   }
 
   if (process.platform === 'win32') {
-    fileUrl = fileUrl.replace(/%5C/g, '/').replace(/\\/g, '/');
+    fileUrl = fileUrl.replace(/%5C/g, '/');
   }
 
   return fileUrl;
--- a/lib/src/utils.ts
+++ b/lib/src/utils.ts
@@ -86,18 +86,28 @@ export function valueError(message: string, name?: string): Error {
   return Error(name ? `$${name}: ${message}.` : `${message}.`);
 }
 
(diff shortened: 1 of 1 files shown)


lib/src/utils.ts
```diff
@@ -86,18 +86,28 @@ export function valueError(message: string, name?: string): Error {
...
+  let fileUrl = encodeURI(path).replace(/[#?]/g, encodeURIComponent);
+
+  if (unsafePathToFileURL) {
+    fileUrl = fileUrl.replace(/%(5B|5D|5E|7C)/g, decodeURIComponent);
+  } else {
+    fileUrl = fileUrl.replace(/~/g, '%7E');
+  }
+
+  if (process.platform === 'win32') {
+    fileUrl = fileUrl.replace(/%5C/g, '/').replace(/\\/g, '/');
```
nex3, line 107: This is changing the behavior even for previous Node versions. Previously we _always_ replaced backslash with forward slash, now we're only doing it on Windows. Is that intentional?

ntkme: Yes. On windows `fileURLToPath` on linux has `/` is escaped as `%5C`, and in the previous implementation you can see that only on windows it's reverting the escape of `%5C`, meaning `.replace(/\\/g, '/')` on linux actually did nothing.

nex3 (approved):

lib/src/utils.ts
ntkme: In fact, `.replace(/\\/g, '/')` seems to be unnecessary...

PR #355, opened 2025-01-10
Remove unneeded `del` development dependency

clydin:
Node.js 16 is the minimum supported version as per the package.json and this version contains the `rmSync` filesystem function which can replace the `del` package usage.


nex3:
Thanks!

PR #356, opened 2025-01-10
Update `tar` development dependency to 7.4.3

clydin:
The latest major version of `tar` contains type definitions within the package itself. This removes the need to install the separate `@types/tar` package.


nex3:
It looks like this is breaking the typings used by minizlib.

Goodwine:
I believe this is broken because the owner of `minizlib` needs to approve and merge this PR:
https://github.com/isaacs/minizlib/pull/31

clydin:
I was looking at opening a PR to enable `esModuleInterop` actually. That would solve this issue and other future subtle ESM/CommonJS interoperability issues which will become more common as packages move to ESM only.

Goodwine:
I believe we can do this now. There is a failing test which I believe to be unrelated to the upgrade tho.
Closing in favor of https://github.com/sass/embedded-host-node/pull/311

PR #357, opened 2025-01-10
Remove unused `minipass` development dependency

clydin:
The `minipass` dependency does not appear to be used within the repository and has now been removed from the `package.json`.


nex3:
Thanks!

Issue #358, opened 2025-01-22
SassBoolean should be a class


nex3:
> Hello,
> 
> I'm trying to migrate a script from `sass` to `sass-embedded` and can't use `SassBoolean` anymore.
> 
> 
> 
> The script tries to identify the type with:
> 
> ```typescript
> 
> if (value instanceof SassBoolean) {
> 
>   (...)
> 
> }
> 
> ```
> 
> 
> 
> At runtime, the following error is thrown:
> 
> `Error: TypeError: Right-hand side of 'instanceof' is not an object`
> 
> 
> 
> I realized `SassBoolean` is an interface and not a class:
> 
> https://github.com/sass/embedded-host-node/blob/248437e15c67a352a9971c8cb3e9addbd8afc107/lib/src/value/boolean.ts#L15
> 
> 
> 
> Therefore, it is not present here in the compiled `index.js` along sassFalse and sassTrue:
> 
> ```javascript
> 
> var boolean_2 = require("./src/value/boolean");
> 
> Object.defineProperty(exports, "sassFalse", { enumerable: true, get: function () { return boolean_2.sassFalse; } });
> 
> Object.defineProperty(exports, "sassTrue", { enumerable: true, get: function () { return boolean_2.sassTrue; } });
> 
> ```
> 
> 
> 
> Should one use `sassFalse` and `sassTrue` instead ?
> 
> As is, `SassBoolean` still breaks usage of `sass-embedded` as drop-in replacement, as mentioned by @rvock. 

 _Originally posted by @guillaumerochelle in [#337](https://github.com/sass/embedded-host-node/issues/337#issuecomment-2593192407)_

[The spec](https://github.com/sass/sass/blob/main/spec/js-api/value/boolean.d.ts.md#sassboolean) defines `SassBoolean` as a class, so this is indeed a bug in the Embedded Host's implementation of the Sass API.

Goodwine:
I'm reading through the code, it seems that `SassBoolean` was left as an interface so that it could not be called with weird things like `new SassBoolean(123 as boolean)`.

To preserve that intent, I think we can simply replace the interface with an abstract class, and then the use case posted by @guillaumerochelle should work, I think

nex3:
The actual typescript annotations that clients see are the ones from the spec directory, so there's no type-checking value in having the implementation have different types.

Goodwine:
Right, this PR should fix that - https://github.com/sass/embedded-host-node/pull/359

[...]

PR #359, opened 2025-01-24
Fix Sass Boolean, make it a class to match spec

Goodwine:
Fixes https://github.com/sass/embedded-host-node/issues/358

See https://github.com/sass/sass-spec/pull/2045


nex3 (approved):
Don't forget to add a changelog entry for this to dart-sass.

PR #363, opened 2025-02-27
Update description for sync-message-port in README.md 

ntkme:
I found that the description for sync-message-port is also outdated, aside from the link being outdate in #362. This PR remove the outdated sentence.


nex3 (approved):
Thanks!

PR #365, opened 2025-03-06
Fix potential Type Error

alexandertrefz:
If the legacy compiler is used, and the file contains 2 `@use` or `@forward` (etc.) clauses, the compiler throws a Type Error, because `prev` is undefined and thus accessing `prev.url` is a Type Error, instead of handling the situation correctly.

### DIFF
--- a/lib/src/legacy/importer.ts
+++ b/lib/src/legacy/importer.ts
@@ -168,7 +168,7 @@ export class LegacyImporterWrapper<sync extends 'sync' | 'async'>
 
     const prev = this.prev[this.prev.length - 1];
     return thenOr(
-      thenOr(this.invokeCallbacks(url, prev.url, options), result => {
+      thenOr(this.invokeCallbacks(url, prev?.url, options), result => {
         if (result instanceof Error) throw result;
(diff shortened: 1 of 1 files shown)


nex3:
Thanks for the fix! Can you add a test case [to the sass-spec repo](https://github.com/sass/sass-spec/blob/main/js-api-spec/legacy/importer.node.test.ts) that reproduces the error and is fixed by this change?

alexandertrefz:
I tried and could not manage to reproduce the issue. It turns out a misbehaving custom importer created the issue on our end.

I would still recommend merging this as a defensive coding precaution, but this doesn't fix a concrete bug.

Apologies for the confusion.

nex3:
If this is something that came from a malformed importer, the best way to handle that would be to throw an error as soon as we got a bad result, rather than trying to _avoid_ throwing an error as long as possible. But generally speaking, we don't add explicit handling for situations where callers violate the TypeScript types unless the violation is exceptionally easy/common. Otherwise we'd end up with tons of type checks everywhere on all our APIs. The recommended way to catch those errors is through typechecking.

PR #367, opened 2025-03-11
Fix passing variables across compilations

ntkme:
This PR addresses https://github.com/sass/dart-sass/issues/2542 for the embedded host side by porting https://github.com/sass-contrib/sass-embedded-host-ruby/compare/c05a8f54a1eaddd70cd7d6278cf111c938a4e93f...0378284e42b304409ef48ce7ef9c9e4dae48a77c.

https://github.com/sass/dart-sass/pull/2544
https://github.com/sass/sass-spec/pull/2053

### DIFF
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -18,6 +18,10 @@ import {Value} from './value';
  * execute them.
  */
 export class FunctionRegistry<sync extends 'sync' | 'async'> {
+  /**
+   * The globally unique identifier of the current compilation used for tracking
+   * the ownership of CompilerFunction and CompilerMixin objects.
+   */
   public readonly compileContext = Symbol();
   private readonly functionsByName = new Map<string, CustomFunction<sync>>();
   private readonly functionsById = new Map<number, CustomFunction<sync>>();
--- a/lib/src/function-registry.ts
+++ b/lib/src/function-registry.ts
@@ -18,6 +18,7 @@ import {Value} from './value';
  * execute them.
  */
 export class FunctionRegistry<sync extends 'sync' | 'async'> {
+  public readonly compileContext = Symbol();
   private readonly functionsByName = new Map<string, CustomFunction<sync>>();
   private readonly functionsById = new Map<number, CustomFunction<sync>>();
   private readonly idsByFunction = new Map<CustomFunction<sync>, number>();
--- a/lib/src/protofier.ts
+++ b/lib/src/protofier.ts
@@ -44,7 +44,7 @@ export class Protofier {
   get accessedArgumentLists(): number[] {
     return this.argumentLists
       .filter(list => list.keywordsAccessed)
-      .map(list => list.id);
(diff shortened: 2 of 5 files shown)


lib/src/function-registry.ts
```diff
@@ -18,6 +18,7 @@ import {Value} from './value';
  * execute them.
  */
 export class FunctionRegistry<sync extends 'sync' | 'async'> {
+  public readonly compileContext = Symbol();
```
nex3, line 21: Please document this.

lib/src/value/argument-list.ts
```diff
@@ -19,7 +19,18 @@ export class SassArgumentList extends SassList {
    * part of the package's public API and should not be accessed by user code.
    * It may be renamed or removed without warning in the future.
    */
-  readonly id: number;
+  readonly id: number | undefined;
```
nex3, line 22: Document what it means for this to be `undefined`.

lib/src/protofier.ts
```diff
@@ -44,7 +44,7 @@ export class Protofier {
   get accessedArgumentLists(): number[] {
     return this.argumentLists
       .filter(list => list.keywordsAccessed)
-      .map(list => list.id);
+      .map(list => list.id as number);
```
nex3, line 47:
```suggestion
      .map(list => list.id!);
```

lib/src/protofier.ts
```diff
@@ -85,15 +85,23 @@ export class Protofier {
...
-        id: value.id,
-        separator: this.protofySeparator(value.separator),
-        contents: value.asList.map(element => this.protofy(element)).toArray(),
-      });
-      for (const [key, mapValue] of value.keywordsInternal) {
-        list.keywords[key] = this.protofy(mapValue);
+      if (value.compileContext === this.functions.compileContext) {
+        const list = create(proto.Value_ArgumentListSchema, {
+          id: value.id,
+        });
```
nex3, line 91:
```suggestion
        const list = create(proto.Value_ArgumentListSchema, {id: value.id});
```

nex3 (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sass-embedded",
   "version": "1.87.0",
-  "protocol-version": "3.1.0",
+  "protocol-version": "3.2.0",
```
Goodwine, line 4: I believe the repo is currently in a broken state because https://github.com/sass/sass/pull/4073 was submitted bumping the protocol version, and the repo expects this to match.

Goodwine: fix: https://github.com/sass/embedded-host-node/pull/372

lib/src/value/argument-list.ts
ntkme: Updated doc string that `undefined` means host constructed argument list. On the object level we don't really care about special value 0 as that's handled by protofier, in which the id default to 0 if unset.

nex3 (approved):

PR #405, opened 2025-12-12
Add support for the display-p3-linear color space

nex3:
See sass/sass#4167
See sass/sass-spec#2095
See sass/dart-sass#2703

### DIFF
--- a/lib/src/value/color.ts
+++ b/lib/src/value/color.ts
@@ -45,6 +45,7 @@ type ChannelNameLch = 'lightness' | 'chroma' | 'hue' | 'alpha';
 type ColorSpaceRgb =
   | 'a98-rgb'
   | 'display-p3'
+  | 'display-p3-linear'
   | 'prophoto-rgb'
   | 'rec2020'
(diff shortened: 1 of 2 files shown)


lib/src/value/color.ts
```diff
@@ -154,6 +155,8 @@ function encodeSpaceForColorJs(space?: KnownColorSpace): string | undefined {
       return 'a98rgb';
     case 'display-p3':
       return 'p3';
+    case 'display-p3-linear':
```
Goodwine: Nit: Do we have [exhaustive checks](https://typescript-eslint.io/rules/switch-exhaustiveness-check/) enabled in eslint?

It doesn't seem like we do so explicitly, it's not enabled by default by eslint, and I couldn't tell whether `gts` enabled it 🤔 

If not, it may be a good idea to do so, or at the very least to have a `const _never: never = value` in a default case for the compiler to warn about it.

nex3: Looks like we do have them enabled.

PR #411, opened 2025-12-27
Enhance sourceMapIncludeSources option

ntkme:
https://github.com/sass/sass/pull/4170
https://github.com/sass/sass-spec/pull/2097
https://github.com/sass/dart-sass/pull/2713

### DIFF
--- a/lib/src/compile.ts
+++ b/lib/src/compile.ts
@@ -5,7 +5,12 @@
 import {initAsyncCompiler} from './compiler/async';
 import {OptionsWithLegacy, StringOptionsWithLegacy} from './compiler/utils';
 import {initCompiler} from './compiler/sync';
-import {CompileResult} from './vendor/sass';
+import {CompileResult, SourceMapIncludeSources} from './vendor/sass';
+import {
+  DeprecationOptions,
+  deprecations,
+  warnForHostSideDeprecation,
+} from './deprecations';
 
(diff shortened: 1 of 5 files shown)


lib/src/compile.ts
```diff
@@ -49,10 +69,34 @@ export async function compileStringAsync(
...
+  sourceMapIncludeSources?: SourceMapIncludeSources | boolean,
+  legacy?: boolean,
+  options?: DeprecationOptions,
+): void {
+  if (
+    legacy !== true &&
+    (sourceMapIncludeSources === true || sourceMapIncludeSources === false)
+  ) {
+    warnForHostSideDeprecation(
+      'Passing boolean value for Options.sourceMapIncludeSources is ' +
```
nex3: Nit:
```suggestion
      'Passing a boolean value for Options.sourceMapIncludeSources is ' +
```

lib/src/compile.ts
```diff
@@ -49,10 +69,34 @@ export async function compileStringAsync(
...
+        'deprecated and will be removed in Dart Sass 2.0.0.\n\n' +
+        'More info: https://sass-lang.com/d/source-map-include-sources-boolean',
```
nex3: It's probably worth including a suggestion here for either `'always'` or `'never'` to preserve the existing behavior.

nex3 (approved):

